PR #57, opened 2019-05-21
pared down schema and corresponding tests

### LINKED ISSUE #55: pare down schema to 1.0
based on agreed scope in [PRC document](https://docs.google.com/document/d/1tbGmw7cb1kozX3gwI_UdbsnS5Lf8BkcjVO9KdN96fTE/edit#)

ahwagner:
closes #55

* Also made a few typo corrections and cleanup

### DIFF
--- a/tests/utils.yaml
+++ b/tests/utils.yaml
@@ -1,7 +1,7 @@
 # Validation tests for utilities
 # see notebooks/appendices/Validation%20Tests.ipynb
 
-get_vmc_sequence_identifier:
+get_vr_sequence_identifier:
   -
     in:
       identifier: "RefSeq:NC_000019.10"
--- a/schema/vr.json
+++ b/schema/vr.json
@@ -36,90 +36,6 @@
          },
          "type": "object"
       },
-      "Bundle": {
-         "additionalProperties": false,
-         "properties": {
-            "alleles": {
-               "default": {},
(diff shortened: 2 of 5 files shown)


tests/utils.yaml
```diff
@@ -1,7 +1,7 @@
 # Validation tests for utilities
 # see notebooks/appendices/Validation%20Tests.ipynb
 
-get_vmc_sequence_identifier:
```
reece: This key is a function name, so needs to be changed in both places.

reece:
I think we agreed to delay this paring down as long as possible. What is the need to pare down now?
It's helpful (for me) to have the fuller vision in place when coding. 
I agree with actual edits.
I'll pick up convo on slack to agree on a paring date.

tests/utils.yaml
ahwagner: Agreed. These notebooks have been moved to vr-python, though, so I've created a separate issue there to track this.

https://github.com/ga4gh/vr-python/issues/23

ahwagner:
I think we want to get a release candidate out to the VR group for review, and that candidate should have the schema as we envision it ASAP. I believe that our branch strategy is to branch on release, so I'm proposing these changes to master now, and branched the 0.1 version to its own branch for reference while developing.

PR #78, opened 2019-06-05
Fixed some typos, use of RFC 2119 and links

cyenyxe:
Some typos fixed, and added use of RFC 2119 in a couple of places. The link to the `Interbase Interval Tests` is now fixed, but it points to an archived file; it should be decided whether to keep it.

In addition to these small changes, I have some suggestions or questions from the [terms and model index page](https://vr-spec.readthedocs.io/en/latest/specification/terms_and_model/index.html):

* Receiving systems SHOULD NOT persist Ids from remote sources. Instead, Identifiers (below) SHOULD be used for communication between systems. -> MUST (NOT) instead of SHOULD (NOT)?
* Use of 'MUST' instead of 'SHALL' could remove the ambiguity that word has (can be a suggestion or a command)

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -1,7 +1,7 @@
 GA4GH Variation Representation Specification
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
-The Variation Representation Specification (VR-Spec) is standard
+The Variation Representation Specification (VR-Spec) is a standard
 developed by the Global Alliance for Genomic Health to facilitate and
 improve sharing of genetic information.  The Specification consists of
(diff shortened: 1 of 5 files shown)


cyenyxe:
Fantastic! What do you think about the couple of questions raised in my first comment, should I address them or leave them be?

ahwagner:
I'm not sure on bullet 1; there may have been a reason this was written as SHOULD NOT; perhaps there may be implementations that wish to for reasons we haven't considered? Anyway, I agree with you, but we should get feedback from @rrfreimuth @larrybabb and/or @reece first.

I agree with you on point 2, please replace any instances you find (though I'm not sure there are any?)

Issue #173, opened 2019-12-26
Allow `Allele.location` to be inlined object or referenced object


reece:
`Allele.location` is currently required to be a Location object.

When serializing Alleles, the `Location` object is replaced by the digest _within an intermediate Python dictionary, not within object itself_.

There is a desire to formally represent Variation Sets using both inline objects and references. However, this would be inconsistent with the current `Allele.location` behavior, which permits only objects.

Before implementing variation sets by id, this issue will allow `Allele.location` properties to be CURIE identifiers (in addition to Location objects, as currently supported).

PR #174, opened 2019-12-29
#173: Allow Allele.location to be a CURIE

reece:
This minor change allows a Allele.location property to be an object (as-is) or a CURIE (new). See #173 for elaboration.

vr-python has minor updates to facilitate this change, which I'll commit separately.

### DIFF
--- a/schema/vr.json
+++ b/schema/vr.json
@@ -9,7 +9,14 @@
                "$ref": "#/definitions/CURIE"
             },
             "location": {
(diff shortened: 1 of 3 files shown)


reece:
It was an accidental byproduct of some testing, but doesn't matter.
I do tend to alphabetize when order doesn't otherwise matter.
So it's a bit of accidental style. :-)

Issue #109, opened 2019-06-10
'Allele' page feedback


mbrush:
https://vr-spec.readthedocs.io/en/latest/specification/terms_and_model/variation.html#allele

1. In the biological def: 

> “allele” refers to a set of specific changes within a Sequence. 

I am not sure I agree with this - but also not sure I understand what you mean by it.  If you mean to say that a common meaning for the term 'allele' is to refer to a **set** of discrete/discontinuous changes, I would disagree.  I think that 'allele' is generally used to refer to a contiguous span of sequence - which may include a single change, or multiple changes.  But the allele itself is a single, located, continuous span of sequence - not a set.

If however you mean that that "allele" refers to **one of a set of** specific variations of a sequence at a particular location, as per the Sequence Ontology definition, then I think this is ok, but should be clarified.

---------------

2. In the IM table, can you use SequenceState as the 'Type' for the state attribute? This would be more consistent with language in the doc that refers to the notion of a 'Sequence' as a component of an allele.  And I don’t think the other planned State subtypes (CNV state and  State Rule) are used to define alleles.

------------

3. For the text: 
> "The implementation MAY infer the Sequence by location_id and the Sequence State type, and ensure compatibility between them. This behavior is not included in the specification."

I don’t really follow your meaning here - in particular what "Sequence" means here (the Sequence State, or the reference sequence?)

------------

4. In the 'Notes' subsection, the following bullet repeats basically the same thing as written in the 'Implementation guidance' subsection:  
> "Alleles may have multiple representations on the same Sequence type due to shifting (aka shuffling, normalization). Fully resolving such equivalences is beyond the scope of the first phase of this specification."

------------

5. Can you link to the example code for Allele in the jupyter notebook, as you did for State (where you write "See example usage in the reference implementation documentation")?

larrybabb:
@mbrush replying to...
2.  This is the discussion that was raised in the VR call today. Folks felt like they didn't want to be too precise with the typing of Allele.state and Allele.location to SequenceState and SequenceLocation, respectively.  The allele.type attribute is meant to be the type of the containing class, in this case 'Allele'.  We would have to change the "type" of state to SequenceState and likely the "type" of location to SequenceLocation to make this more precise as you suggest.  (i happen to agree with you, but folks felt like it would be easier to narrow this definition going forward versus expanding it in the event that other State or Location types emerge that may work for Allele. )

reece:
Thanks @mbrush.

[...]

PR #177, opened 2020-01-24
remove "specific" qualifier per concerns in #109

ahwagner:
Currently the computational definition of `Allele` is (emphasis added):

>An Allele is a **specific**, single, and contiguous Sequence at a Location (Abstract Class). Each alternative Sequence may be empty, shorter, longer, or the same length as the interval (e.g., due to one or more indels).

I propose removing "**specific**", due to the nature of Allele in allowing any `State` and `Location` object and the (ironic) ambiguity of how "specific" is meant to constrain those types (e.g. IUPAC ambiguity codes used in `SequenceState.sequence`), if at all.

### DIFF
--- a/docs/source/terms_and_model.rst
+++ b/docs/source/terms_and_model.rst
@@ -435,10 +435,9 @@ None.
 *State* objects are one of two primary components specifying a VR
 :ref:`Allele` (in addition to :ref:`Location`), and the designated
 components for representing change (or non-change) of the features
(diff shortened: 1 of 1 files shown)


ahwagner:
In GKS leads call today, it sounded that there is general support from @larrybabb and @rrfreimuth for removing the use of "specific" in this definition, with the typical uneasy acknowledgement about the ability to express ambiguous states in VRS.

PR #178, opened 2020-01-24
Remove examples of planned states from State definition

ahwagner:
The computational definition of `State` should be more concrete, and I propose we remove some of the planned concepts that ultimately may or may not become `State` classes (which, per our definition of `State`, makes them allowable under `Allele`).


reece:
I agree that speculative future states do not belong in the reference. However, can we keep the link to the planned-states section? That does seem relevant to me.

ahwagner:
I agree! In fact, the proposed change [does not remove the link](https://github.com/ga4gh/vr-spec/pull/178/files#diff-49b92a17ac0ef3b7b9205d6b662720efR441) to the planned states section.

reece:
Doh! I misread. Thanks. Right. Great. Let's merge!

Issue #182, opened 2020-03-02
Digest: probability of collision


rrfreimuth:
In the appendix on [Truncated Digest Collision Analysis](https://vr-spec.readthedocs.io/en/1.0rc/appendices/truncated_digest_collision_analysis.html#collision-analysis), the number of bits given number of sequences (m) and probability of collision (P) is given as (sorry about the lack of formatting):

```
b(m, P) = \log_2{\left(\frac{m^2}{P}\right)} - 1
```

Farther down the page, in the code block that defines the function B, this formula is given:
```
b = math.log2(m / P) - 1
```

I think we're missing an exponent...
```
b = math.log2(m**2 / P) - 1
```

Am I missing something?

reece:
Yikes. I think you're right.

I'll need a few days to get my head back into these weeds to double check.

Either way, thanks for the careful reading!

rrfreimuth:
I'm working on a PR, will submit it shortly.

PR #183, opened 2020-03-02
Patch collision probability code in appendix

rrfreimuth:
Addresses issue #182 

Reece: Please re-generate the table at the bottom of the page using the patched code.  This patch will change the numbers significantly.

### DIFF
--- a/docs/source/appendices/truncated_digest_collision_analysis.rst
+++ b/docs/source/appendices/truncated_digest_collision_analysis.rst
@@ -360,7 +360,7 @@ is approximately equivalent to:
         Assumes m << 2^b.
         
         """
-        b = math.log2(m / P) - 1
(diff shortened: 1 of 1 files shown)


reece:
Good catch Bob. Thank you! The [table is updated on RTD master](https://vr-spec.readthedocs.io/en/master/appendices/truncated_digest_collision_analysis.html). 

Although the # of bits required changed significantly (as % of bits), overall I'm not concerned because I was so wildly conservative previously about bit selection. The GA4GH Truncated Digest is 24 bytes / 192 bits long. So, for example, with b=192 bits and 1e+21 messages, there P value was (erroneously) 1e-30 and is now 1e-15. Although that's very significant relatively, I feel pretty comfortable that many other problems will hit us before this one (which I guess is a kind of optimism).

PR #184, opened 2020-03-20
Update citation of SPDI Paper

RadixSeven:
Citing the peer-reviewed version rather than the preprint.


reece:
I appreciate the PR!

Issue #116, opened 2019-06-24
Implement explicit Variation Set


reece:
VariationSets are essential to ClinGen's goal to represent ClinVar variation.

This issue is to collect input on and implement a static Variation Set defined by explicit enumeration of constituent variation.

reece:
The current direction is this hierarchy:
```
Variation
+- DiscreteVariation
   +- Allele
   +- (Translocation?)
+- AggregateVariation
   +- DiscreteVariationSet
   +- Haplotype (of DiscreteVariation, with constraints)
   +- Genotype (of Haplotypes)
   +- VariationSet (unconstrained set of Variation)
```

reece:
Status: merged master into 116-variation-sets to pick up the changes from #173 . Now working on getting the enref/deref code to work for variation sets. No problems expected.

reece:
Implemented enref/deref in Implemented enref/deref in 152fae1. Need to add tests to vr-python and vr-spec.

Issue #186, opened 2020-04-19
Implement Haplotypes


reece:
There's been a lot of discussion about how to implement haplotypes and genotypes over the years. Alex, Larry, and Reece recently ratified a plan.

The tl;dr is that we'll proceed with the Haplotype and Genotype models from VMC, with the exception that we'll remove the partial option for the completeness flag for now.

We also considered other approaches, including VCF-style phase sets. We chose against this mostly because it merges the concepts of haplotype and genotype into a single structure, which is attractive but precludes well-scoped classes of the individual concepts.

ahwagner:
After concerns raised on last week's call by @rrfreimuth, we have discussed the intent, purpose, and appropriateness of the `completeness` attribute in our `Haplotype` model. In review, we agreed that the use cases that this attribute covered was not worth the additional complexity in this class.

As a result, we've removed `completeness` from `Haplotype`. We currently are considering a preservation of the optional `location` in `Haplotype`, to allow users to explicitly describe the `Location` over which this `Haplotype` is defined. This allows us to construct precise messages of profiled region to the `Haplotype` object, but can also be omitted for simple message passing.

rrfreimuth:
Thanks for the update!

We need to precisely define the semantics of `location` within the Haplotype class, especially if we allow it to be used for anything other than the outer bounds of the outer Alleles that comprise the Haplotype.

We also need to consider how the optionality of that attribute would impact the digest.  We cannot do this without first establishing the meaning of it, however.

reece:
Also see comments in #33, which I'll now close so that this issue consolidates discussion.

PR #188, opened 2020-04-25
#116-patch2: Revise future plans

ahwagner:
This PR revises the future plans section. This section is in great need of update, and it should happen prior to 1.1. The draft starts with an upgraded diagram with the following changes:
- transposed to use more vertical than horizontal space to accommodate ReadTheDocs figure display format
- revised classes to reflect recent discussions of upcoming VRS classes

Once there is agreement on which classes to present in the revised diagram, this PR should also include associated changes to the `future_plans.rst` document.


reece:
There are 3 branches that I imagine merging into master for 1.1: VariationSet (#116), Haplotype and Genotype (#186), and doc-updates (branch, but no issue). 

VariationSet does not depend on the future plans. I strongly agree that future plans needs updating, but I'd like to to be part of doc-updates. (This was what I meant in my comment to @helensch that this section needed work.)

Okay to push this to doc-updates instead?

ahwagner:
Sure, will close this and cherry-pick the commit for `doc-updates` branch.

PR #196, opened 2020-06-12
Doc updates

reece:
Created as draft PR for tracking purposes on the project page. Not ready to merge (I think).


ahwagner:
Thanks @larrybabb. For now I've revised the future schema image to a more compact form of the original, omitting some details as you suggested and updating some of the entries to reflect recent discussions.

reece:
@ahwagner @larrybabb: I think I'm ready to merge this into master. We might make other changes, but we can deal with those individually. What say you?

ahwagner:
@larrybabb I am waiting to do current. That figure change should be made ahead of `1.1`, but waiting until we finalize classes for that release. Now that I think about it, though, the `future` figure will _also_ need to be changed at that time, and (seeing the milestone tag) I don't think we intend to add these doc updates to `1.0`. I might have been premature in making even this change. 🤷

reece:
I'm taking the two thumbs up on the merge proposal to proceed. Stand by... 3.. 2.. 1...

PR #197, opened 2020-06-15
181 cytoband and gene location

reece:
This is a draft PR for CytobandLocation and GeneLocation.

See the [docs at readthedocs](https://vr-spec.readthedocs.io/en/181-cytoband-and-gene-location/) and the [notebook](https://github.com/ga4gh/vr-python/blob/vr-spec-181-cytoband-and-gene-location/notebooks/Feature%20Based%20Locations.ipynb). 

Please provide comments in this PR.

TODO:

- [x] Finish documentation


ahwagner:
In a GKS leads discussion today, we agreed that while it would be nice to externalize the concept of the physical chromosome that `CytobandLocation` describes, that the only suitable representation of these can be found in Wikidata. While the group was split on whether or not to adopt this model using Wikidata concepts, we agreed to keep the `CytobandLocation` as currently modeled for v1.

In support of the group decision, `CytobandLocation` has a 👍  from me.

docs/source/terms_and_model.rst
```diff
@@ -346,6 +346,186 @@ Location for Variation.
...
+     - Description
+   * - _id
+     - :ref:`CURIE`
+     - 0..1
+     - Location Id; MUST be unique within document
+   * - type
+     - string
+     - 1..1
+     - Location type; MUST be set to **`GeneLocation`**
+   * - gene
```
ahwagner, line 479: I think this should be changed to `gene_id` for consistency with `sequence_id` in `SequenceLocation`
```suggestion
   * - gene_id
```

docs/source/terms_and_model.rst
```diff
@@ -346,6 +346,186 @@ Location for Variation.
...
+     - :ref:`CURIE`
+     - 1..1
+     - An id mapping to the :ref:`computed-identifiers` of the
+       external database Sequence containing the sequence to be
+       located.
+
```
ahwagner, line 485: To accommodate the computational definition and use cases in mind, I would add the optional `region` attribute to the GeneLocation class:
```suggestion
   * - region
     - :ref:`Region`
     - 0..1
     - Region of feature on gene specified by gene_id. If the region is omitted, the GeneLocation is the entire conceptual region of a gene.
     
```

docs/source/terms_and_model.rst
```diff
@@ -346,6 +346,186 @@ Location for Variation.
...
+.. _gene_location:
+
+GeneLocation
+$$$$$$$$$$$$$$$$
+
+**Biological definition**
+
+The location of a gene is the imprecise conceptual location of a gene
+feature on a named archetypal chromosome.
```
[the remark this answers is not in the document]
reece: I wrestled with whether GeneLocation had a biological definition at all. I started much the way you did (with a definition of a gene), but ultimately discarded it because defining gene seems out of scope. (After all, we're saying that we're going to delegate this definition to an external org, so by construction it is not ours to define.)

However, I can get behind this definition too. Sold!

reece:
1. I've made these changes so that we can try these out. Overall, agree with the drive to create parallelism of Location subclasses: SequenceLocation is a location *on* a sequence (including entire sequence), GeneLocation is a location *on* a gene (or entire gene), and ~CytobandLocation~ ChromosomeLocation is a location on a chromosome (hmm...). 

1. I want to reraise my concerns about the particular examples you provided. "EGFR exon 20" is a thing humans say, but it's incomplete: a gene does not have an exon (a transcript does). So, if we want to say that, we must include a reference to the transcript at the very least. A similar argument holds for specificity of model used to define a domain.

reece:
This PR is retracted in favor of separate PRs to handle ChromosomeLocations and gene locations (design TBD).

PR #203, opened 2020-06-25
181 chromosomelocation

reece:
See docs at

https://vr-spec.readthedocs.io/en/181-chromosomelocation/terms_and_model.html#chromosomelocation

Please pay particular attention to the "TBD" items at the bottom of Implementation Guidance.


docs/source/terms_and_model.rst
```diff
@@ -346,7 +346,119 @@ Location for Variation.
...
+.. _chromosome_location:
+
+ChromosomeLocation
+$$$$$$$$$$$$$$$$$$$$
+
+**Biological definition**
+
+Imprecise chromosomal locations based on named landmarks assayed
+through chromosomal staining, FISH, recombination studies, or other
+low-resolution methods.
```
ahwagner, line 358: This is fine, though I might have left off the notion of precision / resolution. While correct, it feels superfluous.

docs/source/terms_and_model.rst
```diff
@@ -346,7 +346,119 @@ Location for Variation.
...
+
+**Computational definition**
+
+A ChromosomeLocation is a :ref:`Location` that is defined by a named
+chromosomal band within a species.  Conceptually, a ChromosomeLocation
+represents a family of :ref:`SequenceLocation` on distinct
+:ref:`Sequence` entities.
```
ahwagner, line 365: It took me a minute to understand what was meant by the family of SequenceLocatio on Sequence entites. I get it, but I wonder if it is useful to have this in the definition–I feel like it does not clarify the preceding sentence, and could cause confusion to people who do not already understand the concept being presented here. I would drop that part of the definition.
```suggestion
chromosomal band within a species.
```

docs/source/terms_and_model.rst
```diff
@@ -346,7 +346,119 @@ Location for Variation.
...
+  <http://hgdownload.cse.ucsc.edu/goldenPath/hg38/database/cytoBand.txt.gz>`__,
+  and `bioutils (Python)
+  <https://bioutils.readthedocs.io/en/stable/reference/bioutils.cytobands.html>`__.
+* Design decision: The tuple <species,chromosome name> is used to
+  refer an archetypal chromosome for the species.  WikiData and MESH
+  provide similar definitions (e.g., Human Chr 1 at `WikiData
+  <https://www.wikidata.org/wiki/Q430258>`__ and `MESH
+  <https://meshb.nlm.nih.gov/record/ui?ui=D002878>`__, but don't
+  provide sufficient benefit to merit the added complexity in VRS at
+  this time.
```
ahwagner, line 439: I think we had decided that, if suitable concepts were found, that it _would_ be worth the complexity, but that the issue was about the stability / governance of WikiData concepts (and sparsity of MESH concepts). This design decision should reflect that, and perhaps be added directly in the _Design Decisions_ section instead of the definition.

docs/source/terms_and_model.rst
reece: I agree. Too wordy.

docs/source/terms_and_model.rst
reece: I don't think we agreed that it would be worth the complexity (or at least I didn't), but you're right that we discussed data stability in a crowd-sourced entity and overall coverage in MESH. I'll add it.

docs/source/terms_and_model.rst
```diff
@@ -342,11 +342,122 @@ Location for Variation.
...
+     - 1..1
+     - The symbolic chromosome name
+   * - start
+     - string
+     - 1..1
+     - The chromosomal band name of the start
+   * - end
+     - string
+     - 1..1
+     - The end chromosomal band name; if null, end is presumed to equal `start`.
```
helensch, line 400: From the implementation guide, I thought end  must be specified.

docs/source/terms_and_model.rst
```diff
@@ -342,11 +342,122 @@ Location for Variation.
...
+  and `end` depend on the species.  Therefore, VRS does not constrain
+  these values in the model, except for the guidelines below.
+* `chromosome` MUST be an official primary name from `NCBI Assembly
+  <https://www.ncbi.nlm.nih.gov/assembly>`__.  For Humans, valid
+  chromosome names are 1..22, X, Y.
+* `end` must be specified, even when `start` and `end` are identical.
+* `start` and `end` SHOULD use syntax and ordering conventions
+  appropriate for the species.
+
+  * For humans, ISCN conventions MUST be used. Bands are denoted by
```
helensch, line 423: Is there a link to ISCN conventions available?

docs/source/terms_and_model.rst
```diff
@@ -342,11 +342,122 @@ Location for Variation.
...
+  MUST be `taxonomy`, corresponding to the `NCBI taxonomy prefix at
+  identifiers.org
+  <https://registry.identifiers.org/registry/taxonomy>`__.
+* Valid values for, and syntactic structure of, `chromosome`, `start`,
+  and `end` depend on the species.  Therefore, VRS does not constrain
+  these values in the model, except for the guidelines below.
+* `chromosome` MUST be an official primary name from `NCBI Assembly
+  <https://www.ncbi.nlm.nih.gov/assembly>`__.  For Humans, valid
+  chromosome names are 1..22, X, Y.
+* `end` must be specified, even when `start` and `end` are identical.
```
helensch, line 419: Should must be capitalised?

docs/source/terms_and_model.rst
```diff
@@ -342,11 +342,122 @@ Location for Variation.
...
+* See also the rationale
+  for :ref:`dd-not-using-external-chromosome-declarations`.
+
+
+**Example**
+
+.. parsed-literal::
+
+   {
+     'chr': '11',
```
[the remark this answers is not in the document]
reece: Yes. The attribute was supposed to be `chr`. Fixed.

docs/source/terms_and_model.rst
reece: Yes to both. Fixed.

docs/source/appendices/design_decisions.rst
```diff
@@ -156,3 +156,20 @@ JSON`_ (not to be confused with `OLPC Canonical JSON`_).
...
+
+
+.. _dd-not-using-external-chromosome-declarations:
+
+Not using External Chromosome Declarations
+@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+
+In :ref:`ChromosomeLocation`, the tuple <species,chromosome name>
+refers an archetypal chromosome for the species.  WikiData and MESH
+provide such definitions (e.g., Human Chr 1 at `WikiData
```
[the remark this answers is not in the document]
reece: Yes to both. Fixed.

schema/vr.json
```diff
@@ -34,8 +34,47 @@
...
+         "pattern": "^\\w[^:]*:.+$",
+         "type": "string"
+      },
+      "ChromosomeLocation": {
+         "additionalProperties": false,
+         "properties": {
+            "_id": {
+               "$ref": "#/definitions/CURIE"
+            },
+            "chr": {
```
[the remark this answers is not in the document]
reece: Fixed. I meant the attribute to be `chr`.

docs/source/terms_and_model.rst
reece: Alas, not really. ISCN is paywalled. A pay-to-access standard is anathema to science, but no one asked me. :-/

docs/source/terms_and_model.rst
reece: Good catch. Fixed.

schema/vr.json
```diff
@@ -34,8 +34,47 @@
...
+               "type": "string"
+            },
+            "end": {
```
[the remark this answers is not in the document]
reece: Not anymore. Fixed.

schema/vr.json
```diff
@@ -34,8 +34,47 @@
...
+               "default": "ChromosomeLocation",
+               "enum": [
+                  "ChromosomeLocation"
+               ],
+               "type": "string"
+            }
+         },
+         "required": [
+            "type",
+            "chr",
```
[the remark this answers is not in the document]
reece: → chr

docs/source/terms_and_model.rst
```diff
@@ -342,11 +342,122 @@ Location for Variation.
...
+     - Location Id; MUST be unique within document
+   * - type
+     - string
+     - 1..1
+     - Location type; MUST be set to **'ChromosomeLocation'**
+   * - species
+     - :ref:`CURIE`
+     - 1..1
+     - An external reference to a species taxonomy.  See Implementation Guidance, below.
+   * - chromosome
```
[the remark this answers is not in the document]
reece: → chr

docs/source/terms_and_model.rst
mbaudis: +1 @ahwagner . This also could lead to confusion since different band resolutions exist w/o good standardization (though the ISCN provides references):

    Chromosomal locations based on named landmarks assayed through chromosomal staining, FISH, recombination studies, or other methods referencing banded karyograms.

... is probably better suited.

docs/source/terms_and_model.rst
mbaudis: Hmmm... I know the responsible person at Karger publisher - in fact, they're working on / evaluating an  electronic edition (but not sure how much would be open). Will ask...

docs/source/terms_and_model.rst
rrfreimuth: +1 @mbaudis

docs/source/terms_and_model.rst
```diff
@@ -342,11 +342,122 @@ Location for Variation.
...
+**Implementation guidance**
+
+* ChromosomeLocation is intended to enable the representation of
+  cytogenetic results from karyotyping or low-resolution molecular
+  methods.  Precise :ref:`SequenceLocation` should be preferred when
+  nucleotide-scale location is known.
+* `species` is specified using the NCBI taxonomy.  The CURIE prefix
+  MUST be `taxonomy`, corresponding to the `NCBI taxonomy prefix at
+  identifiers.org
+  <https://registry.identifiers.org/registry/taxonomy>`__.
```
rrfreimuth, line 412: We should also specify the expected content for the reference portion of the CURIE.

```suggestion
  <https://registry.identifiers.org/registry/taxonomy>`__, and the CURIE reference MUST be an NCBI taxonomy identifier (e.g., 9606 for Homo sapiens).
```

docs/source/terms_and_model.rst
```diff
@@ -342,11 +342,122 @@ Location for Variation.
...
+
+
+**Implementation guidance**
+
+* ChromosomeLocation is intended to enable the representation of
+  cytogenetic results from karyotyping or low-resolution molecular
+  methods.  Precise :ref:`SequenceLocation` should be preferred when
+  nucleotide-scale location is known.
+* `species` is specified using the NCBI taxonomy.  The CURIE prefix
+  MUST be `taxonomy`, corresponding to the `NCBI taxonomy prefix at
```
rrfreimuth, line 410: Note that the proposed structure of the CURIE (`taxonomy:<identifier>`) differs from NCBI's suggested syntax for references within manuscripts.  Example for [humans](https://www.ncbi.nlm.nih.gov/Taxonomy/Browser/wwwtax.cgi?mode=Info&id=9606&lvl=3&lin=f&keep=1&srchmode=1&unlock):

> Taxonomy ID: 9606 (for references in articles please use NCBI:txid9606)

In that form, "NCBI" is the prefix and the reference is a concatenation of "txid" and the taxonomy identifier.

I do not think VRS needs to follow the NCBI recommendations for xrefs in manuscripts, but it is something we should be aware of (and possibly mention within the docs as a known divergence).

docs/source/terms_and_model.rst
```diff
@@ -342,11 +342,122 @@ Location for Variation.
...
+  identifiers.org
+  <https://registry.identifiers.org/registry/taxonomy>`__.
+* Valid values for, and syntactic structure of, `chromosome`, `start`,
+  and `end` depend on the species.  Therefore, VRS does not constrain
+  these values in the model, except for the guidelines below.
+* `chromosome` MUST be an official primary name from `NCBI Assembly
+  <https://www.ncbi.nlm.nih.gov/assembly>`__.  For Humans, valid
+  chromosome names are 1..22, X, Y.
```
rrfreimuth, line 418:
```suggestion
  chromosome names are 1..22, X, Y (case-sensitive).
```

melissacline:
I finally grasped what I now understand to be a basic point: Chromosome Location is **not** a specialization of Sequence Location, rather an alternative form of Location for which the start and end are strings rather than an Interval class.  What about a bullet point under Location -> Implementation Guidance to summarize this?  Or maybe move this text from Chromosome Location -> Implementation Guidance?

> ChromosomeLocation is intended to enable the representation of cytogenetic results from karyotyping or low-resolution molecular methods. Precise SequenceLocation should be preferred when nucleotide-scale location is known.

schema/vr.json
```diff
@@ -34,8 +34,47 @@
...
          "description": "A string that refers to an object uniquely.  The lifetime and scope of an id is defined by the sender.\nVR does not impose any contraints on strings used as ids in messages. However, to maximize sharability of data, the VR Specification RECOMMENDS that implementations use [W3C Compact URI (CURIE)](https://www.w3.org/TR/curie/) syntax.\nString CURIEs are represented as `prefix`:`reference` (W3C terminology), but often referred to as `namespace`:`accession` or `namespace`:`local id` colloquially.\nThe VR specification also RECOMMENDS that `prefix` be defined in identifiers.org.\nThe `reference` component is an unconstrained string.\nA CURIE is a URI.  URIs may *locate* objects (i.e., specify where to retrieve them) or *name* objects conceptually.  VR uses CURIEs primarily as a naming mechanism.\nImplementations MAY provide CURIE resolution mechanisms for prefixes to make these objects locatable.\nUsing internal ids in public messages is strongly discouraged.",
-         "example": "ga4gh:GA.01234abcde",
-         "pattern": "^\\w[^:]+:.+$",
+         "example": "ensembl:ENSG00000139618",
+         "pattern": "^\\w[^:]*:.+$",
```
rrfreimuth, line 38: If the prefix is not present, then the delimiting colon should not be required.
```suggestion
         "pattern": "^\\w([^:]+:)?.+$",
```

docs/source/terms_and_model.rst
reece: Okay!

Changed to "Imprecise chromosomal locations based on named landmarks."

And, sorry @mbaudis... I hadn't seen your comments until just now. Thanks for chiming in.

schema/vr.yaml
```diff
@@ -237,5 +267,5 @@ definitions:
       Using internal ids in public messages is strongly discouraged.
 
     type: string
-    pattern: '^\w[^:]+:.+$'
-    example: "ga4gh:GA.01234abcde"
+    pattern: '^\w[^:]*:.+$'
```
rrfreimuth, line 270: See above for comment on the regex, final resolution should be implemented in both places for consistency

docs/source/terms_and_model.rst
```diff
@@ -342,11 +342,122 @@ Location for Variation.
...
+     - 1..1
+     - The symbolic chromosome name
+   * - start
+     - string
+     - 1..1
+     - The start chromosomal band name
+   * - end
+     - string
+     - 1..1
+     - The end chromosomal band name
```
rrfreimuth, line 400: I know we discussed keeping `start` and `end` as separate attributes vs. implementing them as an `Interval`, but I do not remember the rationale for it (I believe it was to keep them separate for simplicity).  For the record, I prefer maintaining parallelism by packaging them as a type of `Interval`, but will not object if I am an outlier on this one.

schema/vr.yaml
mbaudis: +1 - and had already implemented it so in a small mapping service I'll link soonish...

reece:
> I finally grasped what I now understand to be a basic point: Chromosome Location is **not** a specialization of Sequence Location, rather an alternative form of Location for which the start and end are strings rather than an Interval class. What about a bullet point under Location -> Implementation Guidance to summarize this? Or maybe move this text from Chromosome Location -> Implementation Guidance?
> 
> > ChromosomeLocation is intended to enable the representation of cytogenetic results from karyotyping or low-resolution molecular methods. Precise SequenceLocation should be preferred when nucleotide-scale location is known.

@melissacline: How about the "Location (Abstract Class)" section immediately above it? I don't mind adding new text, but I want to make sure that you saw that first.

schema/vr.yaml
rrfreimuth: My proposed change to the regex (above) was incorrect. I misinterpreted the `\w` as leading whitespace, when I believe it represented a "word" character. The regex pattern does not permit leading or trailing white space, but the proposed version (`pattern: '^\w[^:]*:.+$'`) is correct (provided CURIEs really do permit any character in the reference portion, as indicated by the `.+`).

docs/source/terms_and_model.rst
mbaudis: A cytoband is an "interval". Cytogenetic annotations can expand their fuzziness by dropping levels. For chromosome 8, this could be

* q
* q2
* q24
* q24
* q24.2
* q24.21

... by which you've arrived at good banding resolution around _MYC_ (~4Mb).

mbaudis:
@melissacline `ChromosomeLocation` is (IMO) to address an annotation provenance problem: Many analyses have been annotated w/ cytoband coordinates; and those are independent of whatever genome edition is _en vogue_ (sic).

Notes:

* `SequenceLocation` is not necessarily more precise (there are no bounds e.g. on a `SimpleInterval`); though in practice...
* `Location` seems a good conceptual wrapper?!

docs/source/terms_and_model.rst
reece: Ultimately, CURIEs are local decisions that we hope will follow convention. We currently standardize on prefixes registered in identifiers.org, and I think that's preferable to deferring to each data source.

schema/vr.yaml
mbaudis: True; there are few one-character prefixes :-)

reece: I'm fine with permitting 1-char prefixes.

docs/source/terms_and_model.rst
```diff
@@ -342,11 +342,120 @@ Location for Variation.
 **Implementation Guidance**
 
 * Location refers to a position.  Although it MAY imply a sequence,
-  the two concepts are not interchangable, especially when the
+  the two concepts are not interchangeable, especially when the
   location is non-specific (e.g., a range) or symbolic (a gene).
```
bwalsh, line 346: The document states ` Two objects are considered equal if and only if their respective attributes are equal.`.  In the context of ChromosomeLocation, should the document provide guidance / examples  for equality, contains, or other operations?

docs/source/terms_and_model.rst
rrfreimuth: @mbaudis Right... cytobands represent an interval, but they are not currently modeled as an `Interval` (which bundles start/end for the coordinate-based Location)

docs/source/terms_and_model.rst
reece: I personally think that this is already covered for identity with the excerpt you provided, and that more complex comparisons are out of scope.

Would you like to propose text for consideration?

docs/source/terms_and_model.rst
reece: The major reason for not making these a subclass of Interval is that it would open the possibility that people would use a sequence interval on chromosome reference, or vice versa.

A secondary reason is lack of need (IMO). Whereas SequenceLocations really will have different kinds of intervals, I don't think that the same is true here, so the complexity isn't easily justified.

Admittedly, it's not that much complexity to create a ChromosomeInterval class with start and end attributes (strings). Do you want that reconsidered?

reece: TODO: Based on discussion w/Alex and Larry, create NamedInterval in lieu of ChromosomeInterval

docs/source/terms_and_model.rst
bwalsh: Does MIchael's comment above https://github.com/ga4gh/vr-spec/pull/203#discussion_r448606281 apply?  I've looked for a succinct write up of karyotype / chromosome parsing and operations we might link to, but have been unable to find one.

mbaudis: @bwalsh Not sure what/how to answer this, but let's try:

* any cytogenetic band annotation is _unambiguous_ in its meaning; there are no alternative annotations, only higher or lower hierarchical expressions; therefore, cytogenetic "positions" ("bands" in the widest sense up to chromosome level) can be expressed through a tree structure
* any single cytogenetic band represents a "genomic interval", i.e. can be expressed as an instance of a _SimpleInterval_ if mapped to a reference genome using some coordinate system, depending on a given reference genome
  - most commonly using an ~800-900 band mapping, e. g. provided by UCSC
* cytobands cannot be represented as single precise genome positions; genome positions can be referred to by cytobands (lossy)
* cytobands are easy; karyotype annotations aren't

Btw., I'm trying to track these developments with a service - so please settle on a syntax :-)

* [progenetix.org/services/cytomapper?cytoBands=8q21q24.1&assemblyId=GRCh38](https://progenetix.org/services/cytomapper?cytoBands=8q21q24.1&assemblyId=GRCh38)
* [progenetix.org/services/cytomapper/?type=ChromosomeLocation&assemblyId=hg18&chr=17&start=p11&end=q21](https://progenetix.org/services/cytomapper/?type=ChromosomeLocation&assemblyId=hg18&chr=17&start=p11&end=q21)
* [progenetix.org/services/cytomapper/?assemblyId=GRCh38&chroBases=17:800000-24326000](https://progenetix.org/services/cytomapper/?assemblyId=GRCh38&chroBases=17:800000-24326000)

schema/vr.yaml
mbaudis: @reece So identifiers.org doesn't even support searching single-letter ones; me myself and I just tried - no hope for us getting `I:`, I guess.

docs/source/terms_and_model.rst
reece: Done. See #205.

schema/vr.yaml
reece: Did you try :star: ?

Because bioinformatics really needs unicode namespaces to improve interoperability. :-/

PR #207, opened 2020-07-04
181 chromosomelocation

ahwagner:
Revisions to the recently merged #203

### DIFF
--- a/docs/source/terms_and_model.rst
+++ b/docs/source/terms_and_model.rst
@@ -419,7 +419,7 @@ $$$$$$$$$$$$$$$$$$
 
 **Biological definition**
 
-Imprecise chromosomal locations based on named landmarks.
+Chromosomal locations based on named landmarks.
 
 **Computational definition**
 
--- a/docs/source/terms_and_model.rst
+++ b/docs/source/terms_and_model.rst
@@ -369,7 +369,7 @@ A contiguous region specified by named features.
   the symbolic values "cen", "tel", or "ter"). If `start` and `end`
   are on different arms, they SHOULD correspond to the p-arm and q-arm
   locations respectively. If `start` and `end` are on the same arm,
-  `start` SHOULD be the more centromeric position (i.e., with lower
+  `start` MUST be the more centromeric position (i.e., with lower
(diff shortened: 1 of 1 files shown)


docs/source/terms_and_model.rst
```diff
@@ -371,9 +371,6 @@ A contiguous region specified by named features.
   locations respectively. If `start` and `end` are on the same arm,
   `start` SHOULD be the more centromeric position (i.e., with lower
   band and sub-band numbers).
-* NamedIntervals are currently unoriented. Future versions of VRS may
-  introduce conventions or attributes that permit explicit
-  orientation.
```
ahwagner, line 376: Removing this, as it is unclear what is unoriented about this interval class, or how this differs from a SimpleInterval, which is not explicitly unoriented.

docs/source/terms_and_model.rst
```diff
@@ -422,7 +419,7 @@ $$$$$$$$$$$$$$$$$$
 
 **Biological definition**
 
-Imprecise chromosomal locations based on named landmarks.
+Chromosomal locations based on named landmarks.
```
ahwagner, line 422: Made this change to drop "Imprecise", based off of [comments](https://github.com/ga4gh/vr-spec/pull/203#discussion_r447585697) from myself, @mbaudis, and a +1 from @rrfreimuth.

Minor, but as long as I'm making a revision PR...

docs/source/terms_and_model.rst
reece: The goal here was two-fold: 1) dissuade users from creating meaning where none was intended. For example, writing start→end order as telomereic→centromeric and wanting that to imply something. 2) Leave open the possibility that it would be used later.

I'd like to keep some version of this idea. Feel free to propose other wording.

docs/source/terms_and_model.rst
reece: Yep. Sorry I missed this.

docs/source/terms_and_model.rst
reece: reconsidered. Left as-is.

Issue #189, opened 2020-04-25
Update schema figures


ahwagner:
I recently committed (c160106) a new [future schema diagram](https://github.com/ga4gh/vr-spec/blob/c160106028a32d1d5a67a57f4e01952726edd930/docs/source/images/schema-future.png) to the `docs-update` branch. Please review the diagram and provide comments here.

reece:
Done, finally, in 5879aa9.

PR #209, opened 2020-07-13
Closes #186: Implement Haplotype

### LINKED ISSUE #186: Implement Haplotypes

reece:


larrybabb:
The schema and future plan diagrams need to be updated to reflect the content in the doc. I think we may have decided to do this before we release 1.1, so please ignore if that is the case. The schema diagram is missing items and could use better organization. The future plans does have more of the model included but seems to favor illustrating associations to Allele over other classes that have similar associations (e.g. Haplotype.location).

The vrs.yaml file includes a data type Cytoband and references it in the NamedLocation, but there is no mention of this datatype in the spec. Is this intentional? Shouldn't the yaml represent precisely what i in the spec? If not then is there a section related to the yaml and json schema that explains the physical differences?

reece:
Schema figure updates is #189.

@larrybabb You're right about Cytoband. When CytobandLocation was modified, NamedLocation was supposed to use strings for `start` and `end` to permit the general utility implied by its name. I will fix now.

d-cameron:
Sorry for the late comment.

Do we have a location that is appropriate for more complicated examples or edge cases? I'm trying to work out how/if the model can represent a SNV that occurs within a segmental duplication but that's unclear to me at this point.

It seems like it would require nesting of Haplotypes, although that could considered to be in scope of the SV extension work. Proper handling of these sort of variants probably also requires a redefinition of 'phasing' and 'cis' as the two SNV `Allele`s occur on the same chromosome (i.e. 'cis' under the usual definition) but are distal to each other. In my SV phasing work I've found the usual phasing definitions of cis and trans aren't very well suited regions with more than two copies.

PR #213, opened 2020-07-14
Remove Haplotype Location

ahwagner:
We have decided to wait on the use of `Location` in `Haplotype` until consensus is reached on implementation details. This is a forward-compatible decision that will unblock other efforts.

While our documentation was updated to reflect this change, the schema itself was not. This PR makes the corresponding change to the schema.

### DIFF
--- a/schema/vr.json
+++ b/schema/vr.json
@@ -81,17 +81,6 @@
             "_id": {
                "$ref": "#/definitions/CURIE"
             },
-            "location": {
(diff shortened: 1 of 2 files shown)


reece:
I'm totally mystified how this got through. I know I removed it, and I thought at the same time I removed the corresponding text in the docs. Must have blown it on a stash push/pop during a branch switch. Thanks for catching/fixing.

PR #214, opened 2020-07-14
Cyto atlas

ahwagner:
For reviewing changes between cyto atlas and master branch

### DIFF
--- a/schema/vr.yaml
+++ b/schema/vr.yaml
@@ -36,8 +36,8 @@ definitions:
     properties:
       type:
         type: string
-        enum: [ChromosomeLocation]
-        default: ChromosomeLocation
+        enum: [GeneLocation]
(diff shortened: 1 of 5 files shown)


schema/vr.yaml
```diff
@@ -146,9 +223,70 @@ definitions:
...
+        $ref: "#/definitions/CURIE"
+      type:
+        type: string
+        enum: [RelativeAbundance]
+        default: RelativeAbundance
+      molecularVariation:
+        $ref: "#/definitions/MolecularVariation"
+      relativeQuantity:
+        type: string
+        enum: [greater_than, less_than, equal_to]
```
ahwagner, line 285: @larrybabb suggests: We should consider abstracting this to some set of States attached to Abundance directly

Issue #193, opened 2020-04-28
Clarify normalization of reference Alleles


ahwagner:
We should talk about normalized reference Alleles. Our normalization routine trims reference-matching sequence, and we do not discuss what happens in edge cases where the `SeqenceState.sequence` value precisely matches the `sequence` extracted from the `SequenceLocation`. To illustrate:

Scenario 1:
`Ref: AACTAA` (interbase coordinates [0,6] of SQ.123)
`Alt: AACTAA`
Computed ID: `VA.abc123`

Scenario 2:
`Ref: ACTAA` (interbase coordinates [1,6] of SQ.123)
`Alt: ACTAA`
Computed ID: `VA.xyz789` (does not match Scenario 1)

Scenario 3:
`Ref: AACTAA` (interbase coordinates [0,6] of SQ.123)
`Alt: ATCTAA`
Computed ID: `VA.def456`

Scenario 4:
`Ref: ACTAA` (interbase coordinates [1,6] of SQ.123)
`Alt: TCTAA`
Computed ID: `VA.def456` (does match Scenario 3)

In the above, when dealing with a reference-matching `sequence`, each `Allele` returns a different identifier based upon choice of interval start. (See **Scenario 1** compared to **Scenario 2**)

However, if there are reference-differing residues in `sequence`, each `Allele` returns the same identifier regardless of choice of interval start. (See **Scenario 3** compared to **Scenario 4**).

Whether or not this merits attention is unclear, but the ramifications are that differences between interval start coordinates in `SequenceLocation` are meaningful for some `Allele` objects but not others. This is clearly unintended behavior.

## Proposed Solution 1a
One approach is to create an additional normalization condition (step 2) that if the start and end coordinate match, to set them both to a value of 0. This approach is straightforward, but the resulting message is nonsensical.

## Proposed Solution 1b
Same as 1a, but instead of setting to a value of 0, restore the interval to its original bounds. This allows for sensical messages, however leaves the choice of interval (start _and_ stop) meaningful for reference-matched `sequence` but not for reference-mismatched `sequence`.

## Proposed Solution 2
Before trim, store current interval bounds. Remove normalization step 2 (condition). When rolling left or right in normalization, you must roll to _at least_ those coordinates. This makes choice of interval meaningful for all sequences, except when needing to extend beyond the interval for full justification in repetitive regions. I believe this most closely matches the original intent of our normalization.

## Proposed Solution 3
[...]

PR #215, opened 2020-07-20
#193: ref allele normalization

reece:
This PR addresses shortcomings of the current description of normalization. As stated in [this comment on #193](https://github.com/ga4gh/vr-spec/issues/193#issuecomment-660556914), the goals of this PR are:

* Reserve the word "normalization" to mean the canonicalization process for all Variation classes.
* Require normalization prior to creating a computed identifier for all Variation classes.
* Modify current Allele normalization to handle ref agree, as follows:
  * In §2. Condition, add "If the reference and alternate alleles are empty, return the input location and alleles."

The majority of the changes are rendered in https://vr-spec.readthedocs.io/en/193-ref-allele-normalization/impl-guide/normalization.html. 

https://vr-spec.readthedocs.io/en/193-ref-allele-normalization/impl-guide/computed_identifiers.html has a minor change to require normalization for all objects.

### DIFF
--- a/docs/source/impl-guide/normalization.rst
+++ b/docs/source/impl-guide/normalization.rst
@@ -45,38 +45,63 @@ fully justified form unless there is a compelling reason to do
 otherwise.  In addition, Alleles MUST be normalized in order to
 generate :ref:`computed-identifiers`.
 
-The process for fully justifying two alleles (reference sequence and
-alternate sequence) at an interval is outlined below.
+The process for fully justifying an Allele is outlined below.
 
-.. NOTE:: Align step numbers here and example table below
+0. Given an Allele:
 
-1. Trim sequences:
+   a. Let `reference allele sequence` refer to the subsequence at the
+      Allele's SequenceLocation.
+   #. Let `alternate allele sequence` be the sequence in the Allele's
+      State object.
+   #. Let `start` and `end` initially be the `start` and `end` of the
+      Allele's SequenceLocation.
 
-   * Remove suffixes common to all alleles, if any. Decrement
-     the interval end position by the length of the trimmed suffix.
-   * Remove prefixes common to all alleles, if any. Increment
-     the interval start position by the length of the trimmed prefix.
-   * If neither allele is empty, the allele pairs represent a alleles
-     that do not have common prefixes or suffixes.  Normalization is not
-     applicable and the trimmed alleles are returned.
+#. Trim sequences:
 
-2. Determine bounds of ambiguity:
+   a. Remove suffixes common to the `reference allele sequence` and
+      `alternate allele sequence`, if any. Decrement `end` by the
+      length of the trimmed suffix.
+   #. Remove prefixes common to the `reference allele sequence` and
+      `alternate allele sequence`, if any. Increment `start` by the
+      length of the trimmed prefix.
 
-   * Left roll: While the terminal base of all non-empty alleles is
-     equal to the base *prior* to the current position, circularly
-     permute all alleles *rightward* and move the current position
-     *leftward*. When terminating, return `left_roll`, the number
-     of steps rolled leftward.
-   * Right roll: Symmetric case of left roll, returning `right_roll`,
-     the number of steps rolled rightward.
+#. If `reference allele sequence` and `alternate allele sequence`
+   are empty, the input Allele is a reference Allele.  Return the
+   input Allele unmodified.
 
-3. Update position and alleles:
+#. If `reference allele sequence` and `alternate allele sequence` are
+   non-empty, the input Allele has been reduced to a substitution
+   Allele.  Construct and return a new Allele with the current
+   `start`, `end`, and `alternate allele sequence`.
 
-   * To each trimmed allele, prepend the `left_roll` bases prior to the
-     trimmed allele position and append the `right_roll` bases after
-     the trimmed allele position.
-   * Expand the trimmed allele position by decrementing the start by
-     `left_roll` and incrementing the end by `right_roll`.
+   NOTE: The remainings case is that exactly one of `reference allele
+   sequence` or `alternate allele sequence` is empty.  If `reference
+   allele sequence` is empty, the Allele represents an insertion in
+   the reference.  If `alternate allele sequence` is empty, the Allele
+   represents a deletion in the reference.
+
+#. Determine bounds of ambiguity:
+
+   a. Left roll: While the terminal base of all non-empty alleles is
+      equal to the base *prior* to the current position, circularly
+      permute all alleles *rightward* and move the current position
+      *leftward*. When terminating, return `left_roll`, the number of
+      steps rolled leftward.
+   #. Right roll: Symmetric case of left roll, returning `right_roll`,
+      the number of steps rolled rightward.
+
+#. Fully justify the trimmed allele sequences:
+
+   a. To the `reference allele sequence` and `alternate allele
+      sequence`, prepend the `left_roll` bases prior to the trimmed
+      allele position and append the `right_roll` bases after the
+      trimmed allele position.
+   b. Decrement `start` by `left_roll` and increment `end` by
+      `right_roll`.
+
+#. Construct and return a new Allele with the current `start`, `end`,
+   and `alternate allele sequence`.
+
 
 
 .. _normalization-diagram:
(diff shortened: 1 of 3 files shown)


docs/source/impl-guide/normalization.rst
```diff
@@ -3,50 +3,105 @@
...
 Fully-justified normalization expands such ambiguous representation
 over the entire region of ambiguity, resulting in an *unambiguous*
 representation that may be readily compared with other alleles.
 
-VRS RECOMMENDS that Alleles at precise locations are
-normalized to a fully justified form unless there is a compelling
-reason to do otherwise.
+VRS RECOMMENDS that Alleles at precise locations are normalized to a
+fully justified form unless there is a compelling reason to do
+otherwise.  In addition, Alleles MUST be normalized in order to
```
larrybabb, line 45: @reece 

> ..Alleles MUST be normalized in order to generate computed identifiers.

Are we saying that it's not possible to generate computed identifiers from non-normalized alleles or are we saying that generated computed identifiers from non-normalized alleles are not spec compliant?
"

reece: My inclination was to permit identification of non-normalized alleles. Alex wants computed identifiers limited to normalized alleles. I agreed that the it's easier to loosen than restrict over time, we three (I think?) agreed to require normalization now and and consider relaxing that later. (Even writing this, I have some serious reservations about this direction.)

reece:
@larrybabb : I request that we deal with untrimmed sequence as a separate issue because I think it's actually related to the requirement to normalize prior to generating a computed digest. This topic needs more discussion.

docs/source/impl-guide/normalization.rst
```diff
@@ -3,50 +3,105 @@
...
+variation representation into a canonincal form.  Left- and
+right-shuffling are examples of normalization for sequence variants.
+VRS extends this concept to apply to all classes of VRS Variation.
+VRS normalization minimizes a class of ambiguity that impedes
+comparison of variation across systems.
 
-VRS describes a "fully-justified" normalization algorithm
-inspired by NCBI's Variant Overprecision Correction Algorithm [1]_.
+Implementations MUST provide a normalize function that accepts any
+Variation object and returns a normalized Variation.  Guidelines for
```
rrfreimuth, line 14:
```suggestion
Variation object and returns a normalized Variation object.  Guidelines for
```

docs/source/impl-guide/normalization.rst
```diff
@@ -88,38 +113,33 @@ alternate sequence) at an interval is outlined below.
...
-      -  .. math:: TCAGC \Bigl[ \frac{}{AGC} \Bigr] AGCT  ①
-   *  -  2. Condition: One allele must be empty.
+      -  .. math:: TCAGC \Bigl[ \frac{}{AGC} \Bigr] AGCT
 
-            If the reference allele is empty, the allele set represents an insertion in the reference.
-
-            If the alternate allele is empty, the allele set represents a deletion in the reference.
-
-            If neither is true, the allele set represents a substitution, which is not subject to further normalization.
+   *  -  2. & 3. These conditions are False.
```
annalu12, line 136: Formatting for readability: I am scrolling back and forth between the table example. Is the table wide enough to include the descriptions of each normalization step?How about a link out for a side-by-side comparison of process vs. example?

docs/source/impl-guide/normalization.rst
```diff
@@ -3,50 +3,105 @@
...
+General Normalization Rules
+@@@@@@@@@@@@@@@@@@@@@@@@@@@
+
+* VRS normalization functions are homomorphic: That is, the input and
+  output objects are always of the same `type`.
+* Object types that do not have VRS normalization rules are returned
+  as-is.  That is, all types of Variation MUST be supported, even if
+  such objects are unchanged.
+* VRS normalization functions are idempotent: Normalizing a normalized
+  Variation object returns an object with the same data.
```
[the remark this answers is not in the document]
reece: I was avoiding phrases like "the same object" because I believe that implementations should actually return a copy.

Or, more specifically, the function should always operate in-place or on a copy, irrespective of internal conditions.

docs/source/impl-guide/normalization.rst
ahwagner: Yes. We all agreed to REQUIRE normalization (on all VRS objects, not just Alleles) to prevent people from treating normalization as an optional process, and specifically in the case of Alleles this is to prevent implementations from representing insertions / deletions that are left or right shifted (as seen in other variant representation standards).

In my view, VRS should either specify a normalization policy and adhere to it, or should only specify the message structure and leave normalization out-of-scope. Every rule and recommendation added to the specification adds complexity, and that cost should come with the benefit of enforcing / improving interoperability. If there are serious reservations about requiring normalization before generating computed IDs, then we should also be asking ourselves if normalization even belongs in the spec, or if it should be part of a separate policy document.

docs/source/impl-guide/normalization.rst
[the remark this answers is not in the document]
ahwagner: Great catch @rrfreimuth. 👍

docs/source/impl-guide/normalization.rst
```diff
@@ -3,50 +3,105 @@
 Normalization
 !!!!!!!!!!!!!
 
-Certain insertion or deletion alleles may be represented ambiguously
-when using conventional sequence normalization, resulting in
-significant challenges when comparing such alleles.
+In VRS, "normalization" refers to the process of converting a given
+variation representation into a canonincal form.  Left- and
```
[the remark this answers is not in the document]
ahwagner: Agreed, this should refer to all VRS objects and needs clarification.

docs/source/impl-guide/normalization.rst
```diff
@@ -3,50 +3,105 @@
...
+
+
+Allele Normalization
+@@@@@@@@@@@@@@@@@@@@
+
+Certain insertion or deletion alleles may have ambiguous
+representations when using conventional sequence normalization,
+resulting in significant challenges when comparing such alleles.
+
+VRS uses a "fully-justified" normalization algorithm inspired by
```
helensch, line 37: Does the VRS algorithm match NCBI's VOCA? 

The design_decisions.rst "Allele Are Full Justified" section has "VRS require that Alleles be fully justified from the description of the NCBI Variant Overprecision Correction".   

Is there info on normalization decision  (normalizing all VRS objects vs just VRS Variation objects) that should be added to the design decision page.?

docs/source/impl-guide/normalization.rst
helensch, line 7: Typo in canoni**n**cal -> canonical?

docs/source/impl-guide/normalization.rst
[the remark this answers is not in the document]
rrfreimuth: Fair point on "copy". I was trying to emphasize that a normalized object won't be changed (e.g., attribute order scrambled), as that would break normalization.

docs/source/impl-guide/normalization.rst
reece: In a subsequent meeting, we (Larry, Alex, Reece) agreed that normalization needed to have an option to preserve flanking sequence around a variant in order to be consistent with the representation of an allele of reference agreement sequence, but that this optionality undermines a desire to require normalization.

Therefore, we agreed that variation SHOULD be normalized before digesting, and that we would add an explanation in Design Decisions. I'll work on that addition to this PR.

docs/source/impl-guide/normalization.rst
reece: I agree that the web page constraints require flipping back and forth and that it's awkward. I'll spend a small bit of time trying to work out something better.

docs/source/impl-guide/normalization.rst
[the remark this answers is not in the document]
reece: Fixed in 9b4b033d1fcfdc98e4fabb6137f6d4e41ee8450c.

docs/source/impl-guide/normalization.rst
reece: Yes. VOCA source code wasn't (isn't?) available, so I implemented from scratch. Brad Holmes reviewed the source and the results and agreed that they were the same thing.

The intro to normalization now explains that normalization applies to all VRS Variation objects.

docs/source/impl-guide/normalization.rst
reece: Change made in def5cd0.

docs/source/impl-guide/normalization.rst
reece: I don't see an easy way to generate a more concise/readable table with sphinx. PRs or alternative suggestions are welcomed, but I'm going to resolve this for now since it's not significant enough to block merge.

PR #218, opened 2020-08-10
VRS 1.1 closeout

ahwagner:
For community review, week of 8/10/20

### DIFF
--- a/docs/source/terms_and_model.rst
+++ b/docs/source/terms_and_model.rst
@@ -317,9 +317,9 @@ An :ref:`Interval` with a single start and end coordinate.
     }
 
 
-.. _NamedInterval:
+.. _CytobandInterval:
 
-NamedInterval
+CytobandInterval
 $$$$$$$$$$$$$$
 
 **Computational definition**
@@ -341,7 +341,7 @@ A contiguous region specified by named features.
    * - type
      - string
      - 1..1
-     - Interval type; MUST be set to '**NamedInterval**'
+     - Interval type; MUST be set to '**CytobandInterval**'
    * - start
      - string
      - 1..1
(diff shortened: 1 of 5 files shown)


ahwagner:
Open issue: revert TextVariation to Text, deprecate warn, and consider how to introduce TextVariation.

docs/source/terms_and_model.rst
```diff
@@ -353,17 +353,13 @@ A contiguous region specified by named features.
 
 **Implementation guidance**
 
-* `start` and `end` attributes of NamedInterval are intentionally
+* `start` and `end` attributes of CytobandInterval are intentionally
   specified vaguely in order to accommodate a wide variety of
   uses. Examples include named markers on chromosomes, cytogenetic
   bands, and legacy marker names found in older scientific literature.
```
rrfreimuth, line 359: What are legacy marker names? If we are constraining the name of the class to cytoband, we might want to consider updating this list of examples.

docs/source/terms_and_model.rst
```diff
@@ -379,9 +375,49 @@ A contiguous region specified by named features.
...
    }
 
+.. _TextInterval:
+
+TextInterval
+$$$$$$$$$$$$$$
+
+**Computational definition**
+
+A contiguous region specified by named features.
```
rrfreimuth, line 388: This definition seems to make TextInterval a superclass of CytobandInterval, so additional clarity between these classes might be helpful (e.g., explicitly noting TextInterval should not be used to express Cytobands).  Do we have use cases for TextInterval, or is it introduced simply for parallelism?

docs/source/terms_and_model.rst
```diff
@@ -379,9 +375,49 @@ A contiguous region specified by named features.
...
+     - string
+     - 1..1
+     - name of feature end
+
+**Implementation guidance**
+
+* `start` and `end` attributes of TextInterval are intentionally
+  specified vaguely in order to accommodate a wide variety of
+  uses.
+
```
helensch, line 420: An example of a TextInterval as given for SimpleInterval and CytobandInterval would be useful.

docs/source/terms_and_model.rst
```diff
@@ -379,9 +375,49 @@ A contiguous region specified by named features.
    {
      'end': 'q22.3',
      'start': 'q22.2',
-     'type': 'NamedInterval'
+     'type': 'CytobandInterval'
```
helensch, line 378: In other examples, the key and values are enclosed in double quotes e.g "end" : "q22.3"
The ChromosomalLocation example also has single quotes.

docs/source/terms_and_model.rst
reece: My intent was to capture old named markers (usually a format like `D13S45`, but probably lots more). These are merely names of probes that were used for ISH/FISH. However, I just discovered that NCBI dropped the probe database. So, my "legacy" might actually have become "long-forgotten and irrelevant". I'll drop that phrase because it doesn't add anything.

docs/source/terms_and_model.rst
reece: I am struggling a bit with the breadth of TextInterval. It was a bit of a late addition to our thinking that, IMO, we didn't reach consensus on before including here.

Is there an immediate use case in 1.1? If not, I would prefer that we yank it and reconsider in the context of other classes.

ahwagner:
Closing closeout PR in lieu of feature-specific PRs.

PR #219, opened 2020-08-13
VRS 1.1 closeout-rf

rrfreimuth:
"landmarks" => "features" for consistency

### DIFF
--- a/docs/source/terms_and_model.rst
+++ b/docs/source/terms_and_model.rst
@@ -455,7 +455,7 @@ $$$$$$$$$$$$$$$$$$
 
 **Biological definition**
 
-Chromosomal locations based on named landmarks.
+Chromosomal locations based on named features.
 
 **Computational definition**
 


ahwagner:
Thanks for the PR, @rrfreimuth. I think that the use of _landmarks_ in the biological definition was intentional, as it is a precise term from the ISCN nomenclature. Relevant text from section 2.3.1 of the 2016 guidelines follows (emphasis from source text):

> As defined earlier, a **band** is a part of a chromosome clearly distinguishable from adjacent parts by virtue of its lighter or darker staining intensity. The bands are allocated to various regions along the chromosome arms, and the regions are delimited by specific **landmarks**. These are defined as consistent and distinct morphologic features important in identifying chromosomes. Landmarks include the ends of chromosome arms, the centromere, and certain bands. ... A **region** is defined as an area of a chromosome between two adjacent landmarks.

That said, the biological definition we have provided here isn't _quite_ right. For example, 1p34 is a cytoband that is not delimited by landmarks (it is within region 3, which has two delimiting landmarks), only a differential in staining intensity. So we may wish to clarify this to change from "landmarks" => "metaphase chromosomal bands" or similar.

@reece what do you think?

reece:
Because of the back-and-forth of class renaming, the precise intent
meandered between these two meanings.

"landmarks" was definitely intended, but I think "features" works just as
well, and perhaps even better because it admits a slightly broader and more
transparent interpretation.

reece:
This change was made in the 204-misc-doc-updates branch, commit 5dc2464.

PR #229, opened 2020-09-25
Patch version 1.1.1

ahwagner:
Changes:
- fix typo in Allele information model

Additional changes for the 1.1.1 patch should be added to the `1.1.1-patch` branch prior to merging this PR.


reece (approved):

reece:
Since this isn't urgent, should we wait a week to see if there are other warts to correct?

PR #232, opened 2020-10-03
patch 1.1.1 (b)

ahwagner:
Bah. Accidentally branched from `master`, pulling in the README changes for this v1.1 patch. I see no reason to exclude it though, so leaving.

### DIFF
--- a/docs/source/releases.rst
+++ b/docs/source/releases.rst
@@ -22,29 +22,23 @@ Releases
 1.1
 @@@
 
-1.1.2
-#####
-
-This patch version makes the following correction:
-  * Correct styling / indexing of CytobandLocation in restructuredText to match
-    the current Schema and ER Diagram.
-
-And the following clarification:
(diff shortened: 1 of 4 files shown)


docs/source/releases.rst
```diff
@@ -22,9 +22,20 @@ Releases
 1.1
 @@@
 
+1.1.2
```
reece, line 25: We haven't actually released 1.1.1 yet. Please merge this change log entry with 1.1.1.

ahwagner: Err... what constitutes a release? When it is tagged on GH?

The 1.1.1 changes are already live. https://vr-spec.readthedocs.io/en/1.1/releases.html

reece: Yes, it*should* be when tagged.

I realized that these are live already. Next time, we should not merge until we're ready to tag. My bad.

PR #243, opened 2021-01-11
future-compatibility for vrs 1.2+

ahwagner:


reece:
The genesis of this PR was my miscommunication and miscoordination of the vr.json → vrs.json renaming with vrs-python. Instead of making vr-spec accommodate vrs-python, I'd like to shift the fix back to vrs-python. Specifically:

* Keep the intended naming: 1.1 uses vr.json; 1.2 uses vrs.json. Client code has to adapt as part of that support. That means dropping this PR.
* Clarify the intended (but botched by me) correspondence between vrs-python and vr-spec: develop ~ develop, 0.6 ~ 1.1, and (soon) 0.7 ~ 1.2. This is already done in https://github.com/ga4gh/vrs-python README.
* Ensure that vrs-python develop and 0.6 branches correctly correspond with vr-spec branches and that they work. (Done and pushed)
* Release 0.6.3. (Done; see https://pypi.org/project/ga4gh.vrs/0.6.3/)

PR #265, opened 2021-01-24
46 repeat and abundance

reece:
See rendering in https://vrs.ga4gh.org/en/46-repeat-and-abundance/terms_and_model.html.
Changes in other files are insignificant.

Changes:

* Added documentation for LiteralSequence, DerivedSequence, RepeatedSequence, SequenceExpressions
* Added documentation for both Abundance classes
* Renamed Range to IntegerRange. (Seems that we might want a float range later.)
* Deprecated State and SequenceState
* Rearranged variation to use Molecular, Systemic, Other taxonomy
* Rearranged for top-down presentation (i.e., variation → location → sequence expressions → primitive, rather than the inverse order)

Note: I do *not* think this documentation is perfect. However, I do think that this PR substantially addresses the needs of this issue. We can then come back for more polishing.

### DIFF
--- a/docs/source/terms_and_model.rst
+++ b/docs/source/terms_and_model.rst
@@ -414,7 +414,7 @@ $$$$$$$$$$$$$$$$$
 **Biological Definition**
 
 AbsoluteAbundance is the absolute and quantified amount of an entity
-within a genome, cell, or sample.
+within a system, such as a genome, cell, or sample.
 
 **Computational Definition**
 
--- a/docs/source/terms_and_model.rst
+++ b/docs/source/terms_and_model.rst
@@ -418,7 +418,7 @@ within a genome, cell, or sample.
 
 **Computational Definition**
 
-AbsoluteAbundance is represented as a `subject`, which may be an
+AbsoluteAbundance references a `subject`, which may be an
 Allele or Haplotype, or any object identifiable with a CURIE.
 
 **Information Model**
--- a/docs/source/terms_and_model.rst
+++ b/docs/source/terms_and_model.rst
@@ -1,3 +1,9 @@
+.. todo:: Review enums for consistency of presentation in the
+          information model.
+.. todo:: standardize ````_ v. :ref:````
+
+
+
 Terminology & Information Model
 !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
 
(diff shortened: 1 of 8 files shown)


docs/source/terms_and_model.rst
```diff
@@ -216,6 +201,53 @@ amino acid codes.
...
+     - Description
+   * - type
+     - string
+     - 1..1
+     - MUST be "IntegerRange"
+   * - min
+     - int
+     - 0..1
+     - minimum value; inclusive
+   * - end
```
larrybabb: This should be max instead of end.

docs/source/terms_and_model.rst
```diff
@@ -216,6 +201,53 @@ amino acid codes.
...
+     - int
+     - 0..1
+     - maximum value; inclusive
+
+**Implementation Guidance**
+
+* At least one of min or max must be specified.
```
larrybabb: I'm not sure how only one would be specified? If only the min is specified does that mean the max is unbounded and vice versa? I think we should also state that they can be equal and should be when it is a specific quantity. Also we should state that min should never be greater than max and vice versa.

docs/source/terms_and_model.rst
```diff
@@ -1215,6 +1185,196 @@ order. See :ref:`computed-identifiers` for more information.
...
+     - 0..1
+     - Computed Identifier
+   * - type
+     - string
+     - 1..1
+     - MUST be "AbsoluteAbundance"
+   * - subject
+     - :ref:`Allele` | :ref:`Haplotype` | :ref:`CURIE`
+     - 1..1
+     - Subject of the relative abundance statement
```
larrybabb: change relative abundance to absolute abundance.

docs/source/terms_and_model.rst
```diff
@@ -1215,6 +1185,196 @@ order. See :ref:`computed-identifiers` for more information.
...
+   * - amount
+     - :ref:`IntegerRange`
+     - 1..1
+     - At least one of `amount.min` or `amount.max` must be specified
```
larrybabb: see previous comment. I think both should be specified always unless we are going to clarify what it means to leave one end unbounded. I also think we should explicitly state that min and max should be equal when the range is an exact amount.  min=2, max=2 means exactly 2 copies.

docs/source/terms_and_model.rst
```diff
@@ -1215,6 +1185,196 @@ order. See :ref:`computed-identifiers` for more information.
...
+   * - _id
+     - :ref:`CURIE`
+     - 0..1
+     - Computed Identifier
+   * - type
+     - string
+     - 1..1
+     - MUST be "RelativeAbundance"
+   * - subject
+     - :ref:`MolecularVariation` | :ref:`Gene`
```
larrybabb: I don't recall why AbsoluteAbundance.subject is specified differently than RelativeAbundance. Is this intentional? if so, do we specify the rationale?

docs/source/terms_and_model.rst
```diff
@@ -1215,6 +1185,196 @@ order. See :ref:`computed-identifiers` for more information.
...
+     - 1..1
+     - Subject of the relative abundance statement
+   * - amount
```
larrybabb: I apologize if I am missing something, but doesn't the relativeAbundance need to have the ability to specify the base amount it is relative too? gt, lt, etc. need a reference amount to be relevant.

notebooks/develop.ipynb
```diff
@@ -530,8 +530,27 @@
       "{\n",
       "  \"amount\": \"lt\",\n",
```
larrybabb, line 531: in this example the qualitative `lt` is less than what?

schema/vrs.json
```diff
@@ -246,6 +230,39 @@
          },
          "type": "object"
       },
+      "IntegerRange": {
+         "additionalProperties": false,
+         "description": "A (min,max) range of integer values. At least one of min or max must be specified.",
```
larrybabb, line 236: We must be clear what is meant when only one of the (min,max) pair is defined.

larrybabb (changes requested):

docs/source/terms_and_model.rst
reece: Good catch!

docs/source/terms_and_model.rst
reece: Fixed

docs/source/terms_and_model.rst
reece: Done

docs/source/terms_and_model.rst
reece: I added a statement to clarify. Please resolve if it passes muster.

docs/source/terms_and_model.rst
reece: This happened during some debugging. I reverted it so that they now read identically.

docs/source/terms_and_model.rst
reece: You're not missing something, and it's as-intended.

We agreed to hold our nose that we're not specifying any reference. The whole point of the RelativeAbundance statement is to allow us to represent less-than-useful statements like "increased expression of HER2" from the real world. There's an implied and undefined normal.

notebooks/develop.ipynb
reece: See above.

ahwagner (changes requested):
Some initial thoughts. More review to follow.

schema/vrs.json
```diff
@@ -158,8 +155,9 @@
                "$ref": "#/definitions/SequenceLocation"
             },
             "transformation": {
-               "default": null,
+               "default": "none",
                "enum": [
+                  "none",
                   "reverse",
                   "complement",
```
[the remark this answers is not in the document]
reece: Leaving open for a decision re: string or boolean.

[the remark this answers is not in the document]
reece: @ahwagner and @reece agreed to revcomp boolean attribute instead. @reece will change

ahwagner:
Merging to develop, since we're going to be using it for a mega-PR.

PR #273, opened 2021-02-06
Keep issues alive

ahwagner:
Currently "stayin alive" labeled issues are labeled stale and removed. This PR fixes that.


reece:
I was wondering why that wasn't working...

PR #281, opened 2021-03-19
resolves #280

### LINKED ISSUE #280: VRS json-schema breaks online validators
The current JSON Schema for v1.2 breaks online validation tools.

For example, see this Allele validation.

https://www.jsonschemavalidator.net/s/DQXcGJK0

This appears to be caused by the use of `additionalProperties: true` on abstract classes.

Removing this flag from abstract classes fixes this behavior:

https://www.jsonschemavalidator.net/s/qCVtP9Df

But it does not allow for arbitrary added properties, if that flag persists in concrete classes e.g.:

https://www.jsonschemavalidator.net/s/bchtO5lj

ahwagner:


reece (approved):
I didn't actually test, but looks good to me and make sense.

PR #283, opened 2021-03-19
#282: VariationSet in VRS Python

### LINKED ISSUE #282: VRS Variation Set breaking in VRS-Python
Currently, there is an[ in-code comment](https://github.com/ga4gh/vrs/blob/main/schema/vrs.yaml#L48-L51) about how VariationSet is not working in VRS-Python. Rearranging the placement of VariationSet so it is not the last entry in the JSON Schema definitions fixes the problem.

I don't know why this causes a problem, likely a bug in our y2j parser.

ahwagner:
This PR resolves #282.

### DIFF
--- a/schema/vrs.json
+++ b/schema/vrs.json
@@ -1,54 +1,78 @@
 {
    "$schema": "http://json-schema.org/draft-07/schema",
+   "title": "GA4GH-VRS-Definitions",
+   "type": "object",
    "definitions": {
-      "AbsoluteCopyCount": {
-         "additionalProperties": false,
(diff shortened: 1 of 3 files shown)


ahwagner:
I did confirm that this change resolves the error on loading the Variation Set jupyter notebook at ga4gh/vrs-python@25ed6a85ad52f569af57998b399313cc9975c013.

reece:
Hi @ahwagner. So sorry for dragging my feet.

Tried it and it works for me too. Tests pass.

Thanks for hunting this down!

Issue #285, opened 2021-04-21
Use generalized ranges for locations and counts


reece:
**Note: This issue was originally the 1.2 punchlist. That punch list is now represented as [issues for the 1.2 milestone](https://github.com/ga4gh/vrs/milestone/5).**

Consider whether and how to implement generalized ranges for locations and copy counts.

Rational: 
* NestedInterval does not follow a typical interval structure (i.e., it is a tuple of <inner,outer> rather than <start, end> as used elsewhere in VRS
* There is interest in considering a generalized range type for sequence locations and copy counts. 

Also, consider what to borrow from [FHIR Data Types](https://www.hl7.org/fhir/datatypes.html).

ahwagner:
Added a couple of bullets above to resolve https://github.com/ga4gh/vrs/pull/286#pullrequestreview-645017577 discussion and comments from @pnrobinson and @julesjacobsen on the necessity of nested RepeatedSequenceExpression messages.

ahwagner:
- [ ] Replace NestedInterval with new range structures (specifics TBD)
- [ ] Clarify use of None/null
- [ ] Use additionalProperties uniformly
- [ ] Use required attributes uniformly
- [ ] Evaluate CopyNumber subjects (Location vs SequenceExpression)
- [ ] Revisit RepeatedSequenceExpression as subject of RepeatedSequenceExpression

Issue #270, opened 2021-01-31
Align descriptions in Schema with Terminology


ahwagner:
We have made substantial revisions to the terminology for consistency and style. We should update the schema to match computational definitions. I suggest we do this as the last issue ahead of the 1.2 release, as there may be additional changes in response to community review.

ahwagner:
### IN PROGRESS

reece:
Closed in #322 .

PR #286, opened 2021-04-26
RFC: #285 Use generalized ranges for locations and counts

ahwagner:
PR to address punch-list in #285.

Currently handles changes to Numeric. A few additional changes coming prior to advancing to full PR.

### DIFF
--- a/schema/vrs.json
+++ b/schema/vrs.json
@@ -232,6 +232,52 @@
                   }
                ]
             }
+         },
+         "if": {
+            "properties": {
+               "copies": {
+                  "$ref": "#/definitions/Numeric"
+               }
+            }
+         },
+         "then": {
+            "properties": {
+               "copies": {
+                  "properties": {
+                     "value": {
+                        "type": "integer"
+                     },
+                     "comparator": {
+                        "enum": [
+                           "<=",
+                           ">="
+                        ]
+                     }
+                  }
+               }
+            }
+         },
+         "else": {
+            "properties": {
+               "copies": {
+                  "properties": {
+                     "min": {
+                        "properties": {
+                           "value": {
+                              "type": "integer"
+                           }
+                        }
+                     },
+                     "max": {
+                        "properties": {
+                           "value": {
+                              "type": "integer"
+                           }
+                        }
+                     }
+                  }
+               }
+            }
          }
       },
       "Location": {
@@ -344,7 +390,99 @@
                   }
                ]
             }
-         }
+         },
+         "allOf": [
+            {
+               "if": {
+                  "properties": {
+                     "start": {
+                        "$ref": "#/definitions/Numeric"
+                     }
+                  }
+               },
+               "then": {
+                  "properties": {
+                     "start": {
+                        "properties": {
+                           "value": {
+                              "type": "integer"
+                           },
+                           "comparator": {
+                              "enum": [
+                                 "<",
+                                 ">"
+                              ]
+                           }
+                        }
+                     }
+                  }
+               },
+               "else": {
+                  "properties": {
+                     "start": {
+                        "properties": {
+                           "min": {
+                              "properties": {
+                                 "value": {
+                                    "type": "integer"
+                                 }
+                              }
+                           },
+                           "max": {
+                              "properties": {
+                                 "value": {
+                                    "type": "integer"
+                                 }
+                              }
+                           }
+                        }
+                     }
+                  }
+               }
+            },
+            {
+               "if": {
+                  "properties": {
+                     "end": {
+                        "$ref": "#/definitions/Numeric"
+                     }
+                  }
+               },
+               "then": {
+                  "properties": {
+                     "end": {
+                        "properties": {
+                           "value": {
+                              "type": "integer"
+                           }
+                        }
+                     }
+                  }
+               },
+               "else": {
+                  "properties": {
+                     "end": {
+                        "properties": {
+                           "min": {
+                              "properties": {
+                                 "value": {
+                                    "type": "integer"
+                                 }
+                              }
+                           },
+                           "max": {
+                              "properties": {
+                                 "value": {
+                                    "type": "integer"
+                                 }
+                              }
+                           }
+                        }
+                     }
+                  }
+               }
+            }
+         ]
       },
       "CytobandInterval": {
          "description": "A contiguous region specified by chromosomal bands features.",
@@ -521,10 +659,7 @@
                "default": "SimpleNumeric"
             },
             "value": {
-               "type": [
-                  "number",
-                  "integer"
-               ]
+               "type": "number"
             }
          }
       },
@@ -540,16 +675,15 @@
                "default": "ComparatorNumeric"
             },
             "value": {
-               "type": [
-                  "number",
-                  "integer"
-               ]
+               "type": "number"
             },
             "comparator": {
                "type": "string",
                "enum": [
                   "<",
-                  ">"
+                  ">",
+                  "<=",
+                  ">="
                ]
             }
          }
--- a/schema/vrs.json
+++ b/schema/vrs.json
@@ -223,7 +223,14 @@
                ]
             },
             "copies": {
-               "$ref": "#/definitions/AbsoluteCopyCount"
+               "oneOf": [
+                  {
+                     "$ref": "#/definitions/Range"
+                  },
+                  {
+                     "$ref": "#/definitions/Numeric"
+                  }
+               ]
             }
          }
       },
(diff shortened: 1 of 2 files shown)


schema/vrs.json
```diff
@@ -223,7 +223,60 @@
                ]
             },
             "copies": {
-               "$ref": "#/definitions/AbsoluteCopyCount"
+               "oneOf": [
```
reece, line 226: This will be easier to use if it's always a Range, possibly with equal endpoints.

schema/vrs.json
```diff
@@ -496,46 +633,79 @@
             }
          }
       },
-      "Quantity": {
-         "description": "A value representing a multitude or magnitude.",
+      "Numeric": {
```
reece, line 636: I'd prefer to standardize on nouns as type names as we've done in the remainder of the spec. i.e., Number rather than Numeric.

schema/vrs.yaml
```diff
@@ -187,7 +187,34 @@ definitions:
           - $ref: "#/definitions/Location"
           - $ref: "#/definitions/CURIE"
       copies:
-        $ref: "#/definitions/AbsoluteCopyCount"
+        oneOf:
```
reece, line 190: This will be easier to use if it's always a Range, possibly with equal endpoints.

reece, line 190: This will be easier to use if it's always a Range, possibly with equal endpoints.

schema/vrs.yaml
```diff
@@ -410,52 +472,56 @@ definitions:
         $ref: "#/definitions/CURIE"
 
   # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-  # Quantity
+  # Numerics, Comparators, and Ranges
 
-  Quantity:
+  Numeric:
```
reece, line 477: I would prefer that we standardized on nouns (or adjectival forms) as type names as we've done in the remainder of the spec. i.e., Number rather than Numeric, and SimpleNumber rather than SimpleNumeric.

schema/vrs.yaml
```diff
@@ -184,10 +184,37 @@ definitions:
         oneOf:
           - $ref: "#/definitions/MolecularVariation"
           - $ref: "#/definitions/Feature"
-          - $ref: "#/definitions/Location"
+          - $ref: "#/definitions/SequenceExpression"
```
reece, line 187: Please leave Location in until we agree to remove it.

Without Location, how would we refer to CopyNumber of a region? Via a DerivedSequenceExpression at a Location?

reece, line 187: Please leave Location in until we agree to remove it.

Without Location, how would we refer to CopyNumber of a region? Via a DerivedSequenceExpression at a Location?

ahwagner: Yes, via `DerivedSequenceExpression`. All changes in this draft PR are open to discussion; not put here as the final solution, only something to be discussed and resolved for #285.

ahwagner: I'm seeing this same comment twice–resolving this thread as duplicate.

schema/vrs.yaml
```diff
@@ -249,62 +276,87 @@ definitions:
...
+      definition is for an "extent greater than zero".
     oneOf:
+      - $ref: '#/definitions/NumericInterval'
       - $ref: '#/definitions/SimpleInterval'
-      - $ref: '#/definitions/NestedInterval'
     discriminator:
       propertyName: type
 
-  SimpleInterval:
+  NumericInterval:
```
larrybabb, line 290: Assuming we go forward with the Numeric and Range data types being proposed in this PR, we should seriously consider whether to drop Interval and simply embed it SequenceLocation. It doesn't seem like it will ever be used elsewhere. Range is really the lower level general purpose data type that I think SimpleInterval was supposed to cover. Doing this change will simplify the Location data structures ChromosomeLocation with a start and end HumanCytoband type and SequenceLocation with a start and end Numeric|Range types. So this is an example of how these new changes give us back some simplicity in other areas of the model. IMO

schema/vrs.yaml
```diff
@@ -249,62 +276,87 @@ definitions:
...
+                    value:
+                      type: integer
+                max:
+                  properties:
+                    value:
+                      type: integer
+
+  # SimpleInterval has been moved to DEPRECATED section at bottom.
 
   CytobandInterval:
```
larrybabb, line 361: Link this with previous comment about dropping Interval all together. If so, then CytobandInterval would go away and start/end HumanCytoband attributes would be moved directly onto the ChromosomeLocation concept.

schema/vrs.yaml
larrybabb: agreed. I think ComparatorNumber is odd, so we may want to reconsider our options on that name, Ray?  https://www.mathstips.com/point-ray-line-and-line-segment/
So for our sequencelocations we have an interval who's endpoints are Points (SimpleNumber), Rays (ComparatorNumber) and Segments (Range). These non-geometric names work well if we only needed these datatypes for our linear sequences, but we need them to work for amounts as well, so I suppose ComparatorNumber is clearly something that folks will have to learn about before applying - which is probably a good thing.

schema/vrs.yaml
```diff
@@ -410,52 +472,56 @@ definitions:
...
 
   # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-  # Quantity
+  # Numerics, Comparators, and Ranges
 
-  Quantity:
+  Numeric:
     description: >-
-      A value representing a multitude or magnitude.
+      Lorem ipsum
```
larrybabb, line 479: Here's my attempt.... 

> At an abstract level this is something that can represent a single numeric value, can possibly have units defined (future), and can provide an open range of values in a positive or negative direction either inclusive or exclusive of the originating numeric value it is based on.

larrybabb (changes requested):

schema/vrs.yaml
ahwagner: Using `Location` feels off to me and I think that `SequenceExpression` classes provide more extensibility when it comes to systemic copies of sequences. I do not feel strongly about this, but I would like to avoid a situation where we allow either Location OR SequenceExpression, and I see the latter as more extensible.

schema/vrs.yaml
ahwagner: I think your suggestion makes sense. However, we cannot make breaking changes without a major version release and product review. Perhaps an alternative to the `NumericInterval` inside our current `SequenceLocation` is to instead create a simpler SequenceLocation-like class that covers this? Alternatively, run with `NumericInterval` and then refactor to simpler `SequenceLocation` in 2.0?

schema/vrs.yaml
ahwagner: Sounds fine to me. As these are intended to be general-purpose (not just coordinates), I think that leaves us with:

- Number
  - SimpleNumber
  - ComparatorNumber

schema/vrs.yaml
ahwagner, line 479:
```suggestion
An abstract low-level datatype for representing a number of units.
```

schema/vrs.yaml
```diff
@@ -410,52 +472,56 @@ definitions:
...
     discriminator:
       propertyName: type
 
-  AbsoluteCopyCount:
-    description: >-
-      An integer count of copies. Counts are bounded ranges denoted by minimum
-      and maximum possible values. Absolute copy number counts may not be smaller
-      than zero.
+  SimpleNumeric:
+    type: object
```
ahwagner, line 487:
```suggestion
    description: >-
       A simple number.
    type: object
```

schema/vrs.yaml
```diff
@@ -410,52 +472,56 @@ definitions:
...
+    properties:
+      type:
+        type: string
+        enum: [ "SimpleNumeric" ]
+        default: "SimpleNumeric"
+      value:
+        type: number
+
+  ComparatorNumeric:
+    type: object
```
ahwagner, line 498:
```suggestion
    description: >-
        "A number with an open bound indicated by comparator. The bound operator is interpreted as follows: '>' are all numbers greater than and excluding the value, '>=' are all numbers greater than and including the value, '<=' are all numbers less than and including the value, and '<' are all numbers less than and excluding the value."
    type: object
```

schema/vrs.yaml
```diff
@@ -410,52 +472,56 @@ definitions:
...
+        default: "ComparatorNumeric"
+      value:
+        type: number
+      comparator:
+        type: string
+        enum: [ "<", ">", "<=", ">=" ]
+
+  Range:
+    description: >-
+      An inclusive range of Numeric values.
```
ahwagner, line 513:
```suggestion
      An inclusive range of numbers.
```

schema/vrs.yaml
larrybabb: if we go with the later suggestion then we should look forward to how we would refactor SequenceLocation in 2.0. I think we would still need to deprecate and create a new class for the new "simpler" version being suggested now. So, we could take that step now. Unless my assumption is wrong and we can simply reuse the same name for SequenceLocation and change it's structure in 2.0, in which case I would be fine with waiting. It is just a shame to not be able to change this now since there are very few implementations existing so far.

schema/vrs.yaml
reece: Agreed. Will merge.

schema/vrs.yaml
reece: Agreed: use Number rather than Numeric.

schema/vrs.yaml
ahwagner: Recast this as SequenceInterval and do away with abstract class of same name.

schema/vrs.yaml
ahwagner: this is 2.0 dev

ahwagner:
Documentation updates to be handled in #270, after review and merge of other 1.2 features.

PR #292, opened 2021-04-30
closes #289

### LINKED ISSUE #289: Use additionalProperties and required attributes uniformly
The schema used additionalProperties and required attributes inconsistently. That led to their removal. However, move significantly lessens the validation strictness of the schema. These properties should be reintroduced.

~At the same time, clarify use of null. In particular, decide whether null is a valid value, and, if so, how explicit null, default null, additionalProperties, and required attributes are related.~ Also see `null` value discussion occurring in #275.

ahwagner:


reece:
Because this change piggybacked on another PR that was merged first, this branch is superfluous.

PR #296, opened 2021-05-03
Fixes #295: Revert schema changes that cause exceptions in pjs

### LINKED ISSUE #295: python-jsonschema-objects raises NotImplementedError w/recent changes
In vrs-python, I now get:

```
$ make test
...
NotImplementedError: anyOf is not supported as bare property
```

reece:
Two recent changes conflict with python-jsonschema-objects, an essential library used by vrs-python.

#### `NotImplementedError: anyOf is not supported as bare property`

I attempted several workarounds but eventually had to drop the conditional required. 

#### `RecursionError: maximum recursion depth exceeded in comparison`.

Again, no workarounds were found. Apparently, PJS can't support the recursion created by the negated constraint even though it can handle simpler recursion through a parent class.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,3 +25,5 @@ jobs:
       - name: Test with pytest
         run: |
           pytest
+        env:
+          PYTHONPATH: /home/runner/work/vrs/vrs/
--- a/schema/helpers.py
+++ b/schema/helpers.py
@@ -1,0 +1,11 @@
+import logging
+
+_logger = logging.getLogger(__name__)
+
+
+def pjs_filter(yaml_dict):
+    for message_name, message_definition in yaml_dict['definitions'].items():
+        if 'anyOf' in message_definition:
+            _logger.warning(f'Removing anyOf attribute from {message_name} to avoid pjs error.')
+            del message_definition['anyOf']
+    return yaml_dict
--- a/schema/vrs.json
+++ b/schema/vrs.json
@@ -541,7 +541,7 @@
                "type": "integer"
             }
          },
-         "oneOf": [
+         "anyOf": [
             {
                "required": [
                   "type",
--- a/schema/vrs.json
+++ b/schema/vrs.json
@@ -459,16 +459,7 @@
                "default": "RepeatedSequenceExpression"
             },
             "seq_expr": {
-               "allOf": [
-                  {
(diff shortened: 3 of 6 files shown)


ahwagner:
These constraints are valid and helpful for use with other tools, such as our exports to .proto happening in the [y2p branch](https://github.com/ga4gh/vrs/tree/y2p).

I think that we should not strip these schema elements out because they are not supported by a vrs-python dependency. Instead, I think we should make a patch to vrs-python that preprocesses JSON schema documents to remove attributes that are not supported by `python-jsonschema-objects`.

ahwagner (changes requested):
See above comment.

reece:
I disagree. It's important to not break current functionality in order to enable new functionality. Let's please not break working code to address a side project.

ahwagner:
I think there is a separation of concerns here:
- VRS-python dependency does not support a subset of valid JSON Schema
- VRS is written in JSON Schema, and these revisions would drop expected constraints in VRS

This is not about a particular project or implementation; rather, I do not think we should limit our use / expression of JSON Schema because of shortcomings in a downstream product that uses VRS. This can be handled downstream where those conflicts occur. Tools like https://www.jsonschemavalidator.net/ have no trouble with the use of `anyOf` / `allOf` in the schema.

reece:
You and I agree on the ultimate endpoint: complete decoupling of vrs and vrs-python.

However, there *is* an entanglement, and it's a significant one. It's not okay to break the reference implementation for the project, even briefly.

We can reach the shared, desired outcome by either 1) creating a breaking change in vrs, then forcing a fix in vrs-python, or 2) fixing in vrs-python and then introducing a non-breaking change. We've done 1. I'd prefer 2.

larrybabb (commented):
I'm deferring to both of you guys. I don't have a strong feeling either way in regards to how this ends up. If there is a specific point that I can help settle please let me know.

larrybabb (approved):
I agree with this approach unless and until we can show a reasonable use case that requires nested repeating sequences.

ahwagner (approved):
I believe that this PR fixes the new smoke test and addresses it, without losing any specificity in the JSON Schema.

ahwagner:
Nevermind: `oneOf` is interpreted to mean exactly one, not multiple. [This fails validation](https://www.jsonschemavalidator.net/s/vhgeownk).

Issue #275, opened 2021-02-16
Use of optional vs required nullable attributes in schema


reece:
The schema currently uses the `required` attribute inconsistently. It is unclear whether a not-required attribute is equivalent to a nullable attribute.

For this issue, investigate how to make the schema more consistent.

N.B. Reece has a vague recollection that python-jsonschema-objects conflates default values and nullable values during serialization.

ahwagner:
To restate: we are debating whether we want to express nullable attributes as optional, or if they are each required but can accept `NULL` / `None` values.

This conversation is probably deeper than we want to get into ahead of v1.2 release, so we are setting aside as a separate concern.

rrfreimuth:
I think the concepts are separate, so we should base the decision on what semantics we want to convey.  Since both stated options are capable of expressing similar semantics, the decision should be based on the _interpretation_ of what is expressed.  I favor explicit over implicit, so in cases where a missing attribute could be interpreted (or intended to be interpreted) as "null" then I would prefer to remove the optionality so differences in interpretation are minimized.

reece:
I agree with Bob: I prefer the precision of an explicit null over an optional attribute.

ahwagner:
I also agree. Now I'm wondering if we should use explicit NULL or consider other explicit values that hint at their meaning?

larrybabb:
Food for thought https://www.wired.com/story/null-license-plate-landed-one-hacker-ticket-hell/

Just because I've never worked on a production system that intentionally loaded all optional attributes with a special string that meant "intentionally NO value" or "null" doesn't mean that this is wrong, but I am scratching my head to figure out how/why we are in this predicament.  Can we at least cite other reputable software or messaging specs that require the use of a value for attributes that are NULL? I'm sure I'm just missing it - but I can't recall any off the top of my head.

larrybabb:
Another interesting article ... https://www.lucidchart.com/techblog/2015/08/31/the-worst-mistake-of-computer-science/

ahwagner:
Yeah, I have the same reservations as @larrybabb about requiring `some_type` | `NULL` on attributes. But I also think an explicit, required value is a good thing.

rrfreimuth:
Case 1: `{ attrib_1 = "A", attrib_2 = NULL }`
Case 2: `{ attrib_1 = "A" }`

[...]

PR #297, opened 2021-05-04
proposed Range abstract class and subclasses

ahwagner:
I had an idea about how to clarify the semantics and use of the classes that is better-aligned to our previous solution of Range. This PR is open as a proposed solution to #275.

### DIFF
--- a/schema/vrs.json
+++ b/schema/vrs.json
@@ -472,18 +472,6 @@
                "$ref": "#/definitions/Range"
             }
          },
-         "example": {
-            "type": "RepeatedSequenceExpression",
-            "seq_expr": {
-               "type": "LiteralSequenceExpression",
-               "sequence": "CAG"
-            },
-            "count": {
-               "type": "Range",
-               "min": 15,
-               "max": 20
-            }
-         },
          "required": [
             "type",
             "seq_expr",
@@ -523,37 +511,95 @@
          ]
       },
       "Range": {
-         "description": "An inclusive range of integers. A range may be unbounded by omitting either min or max values. Fully unbounded ranges are not permitted (one of min or max MUST be specified).",
+         "description": "A set of classes that provide the possible representations of integer number ranges, from a single integer (Bound) to bounded ranges (ClosedRange) or half-bounded ranges (OpenRange).",
+         "oneOf": [
+            {
+               "$ref": "#/definitions/BoundInteger"
+            },
+            {
+               "$ref": "#/definitions/BoundedRange"
+            },
+            {
+               "$ref": "#/definitions/HalfBoundedRange"
+            }
+         ],
+         "discriminator": {
+            "propertyName": "type"
+         }
+      },
+      "BoundInteger": {
+         "description": "A single integer range.",
          "type": "object",
          "additionalProperties": false,
          "properties": {
             "type": {
                "type": "string",
                "enum": [
-                  "Range"
+                  "BoundInteger"
                ],
-               "default": "Range"
+               "default": "BoundInteger"
             },
-            "min": {
-               "type": "integer"
-            },
-            "max": {
+            "value": {
                "type": "integer"
             }
          },
-         "anyOf": [
-            {
-               "required": [
-                  "type",
-                  "min"
-               ]
+         "required": [
+            "type",
+            "value"
+         ]
+      },
+      "HalfBoundedRange": {
+         "description": "\"A half-open range with an open bound indicated by comparator. The bound operator is interpreted as follows: '>=' are all integers greater than and including the value, '<=' are all integers less than and including the value.\"",
+         "type": "object",
+         "additionalProperties": false,
+         "properties": {
+            "type": {
+               "type": "string",
+               "enum": [
+                  "HalfBoundedRange"
+               ],
+               "default": "HalfBoundedRange"
             },
-            {
-               "required": [
-                  "type",
-                  "max"
+            "bound": {
+               "type": "integer"
+            },
+            "comparator": {
+               "type": "string",
+               "enum": [
+                  "<=",
+                  ">="
                ]
             }
+         },
+         "required": [
+            "type",
+            "bound",
+            "comparator"
+         ]
+      },
+      "BoundedRange": {
+         "description": "A bounded, inclusive range of integers.",
+         "type": "object",
+         "additionalProperties": false,
+         "properties": {
+            "type": {
+               "type": "string",
+               "enum": [
+                  "BoundedRange"
+               ],
+               "default": "BoundedRange"
+            },
+            "min_bound": {
+               "type": "integer"
+            },
+            "max_bound": {
+               "type": "integer"
+            }
+         },
+         "required": [
+            "type",
+            "min_bound",
+            "max_bound"
          ]
       },
       "Sequence": {
--- a/schema/vrs.yaml
+++ b/schema/vrs.yaml
@@ -355,15 +355,6 @@ definitions:
           - $ref: "#/definitions/DerivedSequenceExpression"
       count:
         $ref: "#/definitions/Range"
-    example:
-      type: RepeatedSequenceExpression
-      seq_expr:
-        type: LiteralSequenceExpression
-        sequence: CAG
-      count:
-        type: Range
-        min: 15
-        max: 20
     required: [ "type", "seq_expr", "count" ]
 
   # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
@@ -396,27 +387,68 @@ definitions:
     required: [ "type", "gene_id" ]
 
   # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-  # Range
+  # Numerics, Comparators, and Ranges
 
   Range:
     description: >-
-      An inclusive range of integers. A range may be unbounded by omitting either
-      min or max values. Fully unbounded ranges are not permitted (one of min or
-      max MUST be specified).
+      A set of classes that provide the possible representations of integer number
+      ranges, from a single integer (Bound) to bounded ranges (ClosedRange) or
+      half-bounded ranges (OpenRange).
+    oneOf:
+      - $ref: "#/definitions/BoundInteger"
+      - $ref: "#/definitions/BoundedRange"
+      - $ref: "#/definitions/HalfBoundedRange"
+    discriminator:
+      propertyName: type
+
+  BoundInteger:
+    description: >-
+        A single integer range.
     type: object
     additionalProperties: false
     properties:
       type:
         type: string
-        enum: [ "Range" ]
-        default: "Range"
-      min:
+        enum: [ "BoundInteger" ]
+        default: "BoundInteger"
+      value:
         type: integer
-      max:
+    required: ["type", "value"]
+
+  HalfBoundedRange:
+    description: >-
+      "A half-open range with an open bound indicated by comparator. The bound operator
+      is interpreted as follows: '>=' are all integers greater than and including
+      the value, '<=' are all integers less than and including the value."
+    type: object
+    additionalProperties: false
+    properties:
+      type:
+        type: string
+        enum: [ "HalfBoundedRange" ]
+        default: "HalfBoundedRange"
+      bound:
         type: integer
-    anyOf:
-      - required: [ "type", "min" ]
-      - required: [ "type", "max" ]
+      comparator:
+        type: string
+        enum: [ "<=", ">=" ]
+    required: ["type", "bound", "comparator"]
+
+  BoundedRange:
+    description: >-
+      A bounded, inclusive range of integers.
+    type: object
+    additionalProperties: false
+    properties:
+      type:
+        type: string
+        enum: [ "BoundedRange" ]
+        default: "BoundedRange"
+      min_bound:
+        type: integer
+      max_bound:
+        type: integer
+    required: ["type", "min_bound", "max_bound"]
 
 
   # =============================================================================


schema/vrs.yaml
```diff
@@ -396,27 +387,68 @@ definitions:
...
 
   Range:
     description: >-
-      An inclusive range of integers. A range may be unbounded by omitting either
-      min or max values. Fully unbounded ranges are not permitted (one of min or
-      max MUST be specified).
+      A set of classes that provide the possible representations of integer number
+      ranges, from a single integer (BoundedInteger) to bounded ranges (ClosedRange)
+      or half-bounded ranges (OpenRange).
+    oneOf:
```
larrybabb, line 397: I'll only whine a little here since you are doing such fantastic work. But I'm not a fan of the names.
There's an unnatural tension between calling a single precise value a "kind of" Range just as much as there is calling a min and max range a "kind of" Integer, point, count, etc.. 

I  keep going back to the term "Quantity" which is the amount of something. And while I agree it seems perverse to think of a position on a coordinate system as a Quantity, it does feel okay calling a single fixed value as well as a range of values (open or closed) as a quantity, since quantities conceptually can be precise or ambiguously referenced.

We may simply want to NOT try so hard to reuse this general-purpose data type for both coordinate positions and for quantities (like counts and measures).  While I agree it would be nice to not replicate two very similar patterns it seems like we keep going in circles with this.  I could live with using the term "Quantity" for the position on a coordinate system since it can be construed as the amount of units that the position represents.

Then SimpleQuantity, UnboundedQuantity and Range serve the purposes of both Endpoints on a sequence location as well as  Counts in a copy number abundance concept. And it would set the stage to use these in other measures defined within systems - either coordinate-wise or amount-wise.

But since the term Quantity is a non-started (I assume) let's try something else...

How about this

IntegerValue (abstract class name)
* FixedInteger (or SimpleInteger)
* OpenIntegerRange
* IntegerRange (or ClosedIntegerRange)

I still think the first two (FixedInteger & OpenIntegerRange) could share a common parent since they both have a base-value attribute and the OpenIntegerRange has only the addition of a comparator attribute. But I really don't see the justification, value or logic in having the IntegerRange inherit from the same parent as those two unless one of the two (min or max) is going to be the shared "base-value" that is inherently represented by the ancestor concept and the other is an outer limit. Keeping with the min/max model for Closed Range does mean that both min and max are conceptually identical as the Fixed or Simple Integer.

Again, apologies for the monkey wrench, but I feel like we are so close and I'd like to see if there's any changes in the understanding of our collective that we may all see some of these arguments in a better light now.

Or can counter args be made to help bring me closer to the fold.

schema/vrs.yaml
```diff
@@ -396,27 +387,68 @@ definitions:
...
+      is interpreted as follows: '>=' are all integers greater than and including
+      the value, '<=' are all integers less than and including the value.
+    type: object
+    additionalProperties: false
+    properties:
+      type:
+        type: string
+        enum: [ "HalfBoundedRange" ]
+        default: "HalfBoundedRange"
+      bound:
```
larrybabb, line 430: I think this should be called "value" as well (not bound). Since it is the "value" that is known and will almost certainly be used in similar fashions as the value in the BoundedInteger but with some additional considerations. We may want to consider carrying the "value" attribute thru to the BoundedRange below and simply add an "outerValue" so that we can consider the "value a consistent basis for each of the Integer subtypes and the comparator (in the open range) and the outerValue  in the (bounded range) would be the extensions that these additional "range" classes use to provide the level of ambiguity commensurate with the requirements we've identified.

Value is an attribute that should be on the ancestor class. And all descendants should be able to define themselves relative to a base-value. If not, then they should be considered different classes. This is why the BoundedRange (IMO) needs to either be a separate class which references SimpleInteger (or BoundedInteger above) or it should have the notion of it's base-value and added attributes to represent the "other" limit or bound that satisfies our requirements for Counts and Positions.

larrybabb (changes requested):
See comments. Need to hear back from either/both of you whether any of my commented arguments resonate and warrant further consideration. I will go quietly for now if not, but I will likely need further justification of the current form to be able to bring myself completely on board with it.

julesjacobsen:
Why not keep it simple and since you define range as a fully-closed range, you don't need a `FixedInteger` - that is simply a range `[N, N]` rather than `[N, N+x]`. VCF did us *zero* favours by using `POS` as a shorthand for an interval of 1 and then having to expand it with an `END` in some cases, or requiring clients to count the REF/ALT allele lengths and do come maths to figure out what it all means. Please don't allow the same!

schema/vrs.yaml
julesjacobsen: 'OpenIntegerRange` would imply the interval `(a, b]`, `[a, b)` or `(a, b)` i.e. left-open, right-open or fully-open to someone not reading the docs and it's not immediately obvious from the name either.

schema/vrs.yaml
julesjacobsen: Are you trying to cater for uncertain ranges? Like the `STARTCI` and `ENDCI` in the VCF spec? e.g. trying to say "I know it starts  at position 500, -200, + 300 and ends at about 3000, -450, + 550". The `HalfBoundedRange` '`>=10`' is equivalent to `[10, infinity]` and '`<=10`' = `[1, 10]`, assuming you're using an unsigned, one-based coordinate system.

larrybabb:
@julesjacobsen we may be talking about two separate things. I do agree that a simple Range is a good idea. We are trying to create a general-purpose data type (using classes) so that we can use it for both positions that are endpoints on an interval as well as counts in copy numbers and eventually other types of quantifiable measurements that can be precise and/or ambiguous (open-ended and closed ranges). In doing this, we are challenged by the ability to make this general-purpose type a good fit for all of these cases so it will likely not be ideal for any one. We also have the opportunity to harmonize with the enormous effort and forethought that FHIR has put behind the similar (if not identical) general-purpose types. On one hand I am very compelled to simply leverage FHIR's Quantity, SimpleQuantity and Range data types out of the box. This would have the benefit of having us spend enormous and valuable effort in trying to provide the same depth and understanding that they have put into their message structures (not realistic) and it would also would demonstrate our ability and willingness to harmonize on portions of our models. But on the other hand, we want to own and be responsible for modeling precisely what our constituency wants and not presume FHIR's models are an automatic fit.  Challenges all around.

As far as the VCF issue goes. I agree completely. But our SequenceLocation model will have a "Start" and "End" attribute which will use the Integer class or the OpenRange or BoundedRange classes so that we can convey all the various forms of variation that HGVS provides, which should more than cover VCF and not limit it to only POS.

julesjacobsen:
@larrybabb OK, that makes sense. Does `FixedInteger` still need to exist though? Does it need to be an object to represent a primitive type? Wouldn't `Integer`, `CopyNumber` and `Range` satisfy the requirements?

larrybabb:
There's still a bit of detail that isn't coming thru. While it seems simple, it's not. It's the issue of providing something very simple for the very specific purpose (which is great), but then creating other very simple things for very simple purposes that are slightly similar and thus require a level of duplication and growth in typing and classes. We are doing our best to weight the trade-offs. No one ever said modeling for standards purposes was easy. We are trying not to have to change things every release.

schema/vrs.yaml
```diff
@@ -355,15 +355,6 @@ definitions:
           - $ref: "#/definitions/DerivedSequenceExpression"
       count:
         $ref: "#/definitions/Range"
-    example:
```
reece, line 358: Why did you remove the example?

schema/vrs.yaml
```diff
@@ -396,27 +387,68 @@ definitions:
...
   Range:
     description: >-
-      An inclusive range of integers. A range may be unbounded by omitting either
-      min or max values. Fully unbounded ranges are not permitted (one of min or
-      max MUST be specified).
+      A set of classes that provide the possible representations of integer number
+      ranges, from a single integer (BoundedInteger) to bounded ranges (ClosedRange)
+      or half-bounded ranges (OpenRange).
+    oneOf:
+      - $ref: "#/definitions/BoundedInteger"
```
reece, line 398: An integer doesn't have bounds, so the meaning of BoundedInteger isn't transparent to me.

schema/vrs.yaml
```diff
@@ -396,27 +387,68 @@ definitions:
...
+      - $ref: "#/definitions/BoundedRange"
```
reece, line 399: Boundedness seems like a property of the attributes of the range, not of the range itself.

If each attribute (min, max) has two options (open or closed), then we have, in principle, 4 possible range-like things.

So, I'd prefer to have open/closed be a notion of the point, not the range.

reece (commented):
From our last conversation, I thought we agreed to Range, Number, and ComparatorNumber. Did I misunderstand?

Overall, the proposed classes (BoundedInteger, BoundedRange, HalfBoundedRange) do not seem like the most straightforward implementation of the concepts we've discussed.

schema/vrs.yaml
ahwagner: @larrybabb this is no problem. I will revise to align more closely to our previous work, starting from the names you suggest here. I do agree with @julesjacobsen that the use of `Open` / `Closed` conflicts with the usage of those descriptions [in other contexts](https://en.wikipedia.org/wiki/Interval_(mathematics)), so perhaps we could have:

- `SimpleInteger`
- `HalfBoundedIntegerRange` (instead of `OpenIntegerRange`)
- `BoundedIntegerRange` (instead of `ClosedIntegerRange`)

Since we don't actually have inheritance _per se_ using JSON Schema draft 7, I think we should de-emphasize concerns about attribute inheritance patterns. Instead, I think it would be reasonable for us to create a conceptual grouping to contain all three of these classes (similar to our previous conceptual groupings such as `MolecularVariation`). This grouping could be something along the lines of `IntegerOrRange`, and would serve as a a single reference we could use whenever any of these three classes are acceptable, resulting in:

- `IntegerOrRange`
  - `SimpleInteger`
  - `HalfBoundedIntegerRange`
  - `BoundedIntegerRange`

How does this sound?

schema/vrs.yaml
ahwagner: > I think this should be called "value" as well (not bound). Since it is the "value" that is known and will almost certainly be used in similar fashions as the value in the BoundedInteger but with some additional considerations. We may want to consider carrying the "value" attribute thru to the BoundedRange below and simply add an "outerValue" so that we can consider the "value a consistent basis for each of the Integer subtypes and the comparator (in the open range) and the outerValue in the (bounded range) would be the extensions that these additional "range" classes use to provide the level of ambiguity commensurate with the requirements we've identified.

No problem. I will change this and similar attributes back to `value`.

> Value is an attribute that should be on the ancestor class. And all descendants should be able to define themselves relative to a base-value. If not, then they should be considered different classes. This is why the BoundedRange (IMO) needs to either be a separate class which references SimpleInteger (or BoundedInteger above) or it should have the notion of it's base-value and added attributes to represent the "other" limit or bound that satisfies our requirements for Counts and Positions.

Okay, yes, I agree. I will make these changes.

schema/vrs.yaml
ahwagner: I plan to put this back once we have landed on a solution. This is more of a documentation concern and I do not want to have misleading examples get missed in the final presentation of the schema.

schema/vrs.json
```diff
@@ -588,17 +588,17 @@
             "comparator"
          ]
       },
-      "Range": {
+      "IntegerRange": {
          "description": "A bounded, inclusive range of integers.",
```
larrybabb, line 592: Are we adding any checks/constraints to assure that min is less than max? Also, we may want to state that if min=max that it will be considered equivalent to SimpleInteger.

larrybabb (approved):

schema/vrs.yaml
```diff
@@ -546,8 +546,7 @@ definitions:
     deprecated: true
     description: >-
       DEPRECATED: A SimpleInterval represents a span of sequence. Positions are
-      always represented by contiguous spans using interbase
-      coordinates.
+      always represented by contiguous spans using interbase coordinates.
 
       This class is deprecated. Use SequenceInterval instead.
```
larrybabb, line 551: It is a little sad that we'll be turning around and DEPRECATING SequenceInterval in v2, but I guess we need to draw the line somewhere. sigh.

schema/vrs.json
ahwagner: JSON Schema does not support relative value constraints. But I agree we should add this constraint to our text definition & documentation.

ahwagner:
@reece you good with these revisions?

schema/vrs.yaml
```diff
@@ -396,27 +387,77 @@ definitions:
...
+    description: >-
+      A simple integer value. This is the minimal implementation of an IntegerValue
+      subclass.
     additionalProperties: false
     properties:
       type:
         type: string
-        enum: [ "Range" ]
-        default: "Range"
+        enum: [ "OpenInteger" ]
```
reece, line 420: Class is SimpleInteger by type is OpenInteger

schema/vrs.yaml
```diff
@@ -396,27 +387,77 @@ definitions:
...
+        default: "OpenInteger"
+      value:
+        type: integer
+    required: [ "type", "value" ]
+
+  OpenInteger:
```
reece, line 426: Why "Open"? The enums  are >= and <=, so it's actually closed.

schema/vrs.yaml
reece, line 420: Class is SimpleInteger but type enum says OpenInteger

schema/vrs.yaml
```diff
@@ -396,27 +387,77 @@ definitions:
     required: [ "type", "gene_id" ]
 
   # - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-  # Range
+  # Numerics, Comparators, and Ranges
+
+  IntegerOrRange:
```
reece, line 392: Idea: How about creating a naming convention for internal classes/abstractions that are never instantiated? e.g., _IntegerOrRange.

Note that my preference is to do away with this abstraction entirely, in which case the name suggestion is moot.

schema/vrs.yaml
```diff
@@ -191,7 +191,7 @@ definitions:
           - $ref: "#/definitions/SequenceExpression"
           - $ref: "#/definitions/CURIE"
       copies:
-        $ref: "#/definitions/Range"
+        $ref: "#/definitions/IntegerOrRange"
```
reece, line 194: My preference would be to just inline the oneOf. Why? Because it (IMO) reduces a level of indirection when reading the using class.

That is, the docs for this class will read one of two ways.
1) `copies: IntegerOrRange` (and you have to go see what that means)
2) `copies: Integer | IntegerRange`

The latter is more transparent.

schema/vrs.yaml
```diff
@@ -274,9 +274,9 @@ definitions:
         enum: [ "SequenceInterval" ]
         default: "SequenceInterval"
       start:
-        $ref: "#/definitions/Range"
+        $ref: "#/definitions/IntegerOrRange"
```
reece, line 277: Ditto re: inlining

schema/vrs.yaml
```diff
@@ -396,27 +387,77 @@ definitions:
...
-      max MUST be specified).
+      An abstract class for related integer-based value classes.
     type: object
+    oneOf:
+      - $ref: "#/definitions/SimpleInteger"
+      - $ref: "#/definitions/OpenInteger"
+    discriminator:
+      propertyName: type
+
+  SimpleInteger:
```
reece, line 412: Can we please drop the Simple part? Everytime I read "SimpleInteger" I wanted to put a bag over my head. "Integer" should suffice.

schema/vrs.yaml
```diff
@@ -396,27 +387,77 @@ definitions:
...
+      A set of classes that are used to represent either single integer values
+      (such as SimpleInteger) or integer ranges (such as IntegerRange).
+    oneOf:
+      - $ref: "#/definitions/IntegerValue"
+      - $ref: "#/definitions/IntegerRange"
+    discriminator:
+      propertyName: type
 
-  Range:
+  IntegerValue:
```
reece, line 402: This, too, feels like an unneeded abstraction. I'm skeptical about this class as a concept.

In addition, the name is problematic because there's not obvious linguistic distinction between IntegerValue and SimpleInteger (or Integer with the previous suggestion).

This confusion is borne of the need to name an abstraction that provides little real distinguishing value.

reece (changes requested):

schema/vrs.yaml
ahwagner: Will implement.

schema/vrs.yaml
ahwagner: Okay, agreed. Will remove.

schema/vrs.yaml
ahwagner: Okay, will remove.

schema/vrs.yaml
ahwagner: Agreed.

schema/vrs.yaml
ahwagner: Okay. This concept is directly analogous to a "half-bounded interval", but since we are using interval to mean something else in VRS, I propose we go with `HalfBoundedRange`.

ahwagner:
@reece kindly review at your earliest opportunity.

reece (commented):
I previously wrote a comment that discouraged the use of the various kinds of ranges (open, halfopen, etc) and preference for putting the boundedness in the attribute. I don't see that comment now and am concerned it was lost somehow.

I will have time to write up what I mean after 11am my time.

In short, I prefer the Integer, CompatorInteger, and Range, where Range could use Integer or ComparatorInteger. This puts the notion of open/closed with the endpoint, not with the Range itself.

reece:
@ahwagner , @larrybabb

## Scope

The scope of this discussion is anything involving a continuous range of discrete integer values. Real-valued data are out-of-scope. 

## Terms

* integer (by definition does not include infinity)
* definite range -- a pair of integers
* indefinite range -- a pair of values in which one or both are infinite. Alternatively (almost), a single integer and direction.
* counts -- I mean tandem repeat or copy counts. I assume that we will structure these identically.
 
## Examples/Use cases

- A *definite* span of sequence as interbase coordinates, possibly zero-length, e.g., `[5, 7]`
- An *indefinite* span of sequence as interbase coordinates, e.g., `[2, ∞)` or `(2, ∞)`
- A span of sequence where the start and/or end fall within ranges, e.g., 
- An exact repeat or copy count, e.g., "2 copies"
- A definite range of counts, e.g., "2-to-4 copies, inclusive"

## Proposal

as-is refers to what's in main right now.

- New: DefiniteRange (alt IntegerRange) = <min, max> represents a definite range (structure as-is, with rename)
- New: IndefiniteRange = <value, comparator> (This is the same thing as the ComparatorInteger) but recast as a range
- Change: Range to be be a abstract type that is DefiniteRange | IntegerRange
- Keep: SequenceInterval start and end are Ranges (as-is)
- Keep: SimpleInterval deprecated (as-is)
- Consider: Integer class, so that multi-classed attributes are (e.g.) Integer | DefiniteRange rather than int | DefiniteRange (i.e., so that the attribute value always has a type field)

## Questions

- For IndefiniteRange, should we allow both inclusive and exclusive comparators? That is, should we allow all of [=>, <=, <, >] or just one set of pairs? (My vote is just the inclusive forms.)
-

ahwagner:
@reece:

@larrybabb and I talked this over and we intend to adopt the `DefiniteRange` and `IndefiniteRange` as described. We are going to avoid abstract classes for now (i.e. no `Range`) and instead use inline oneOf as we are doing currently. Since those can be trivially added downstream we can discuss at a later time. We agree with your preference to leave the enumerable as-is, limited to "<=" and ">=".

reece (approved):

reece:
Looks great guys! Thanks for persevering. I'll merge.

Issue #299, opened 2021-05-10
Create a Google Protocol Buffer artefact of the VRS schema


ahwagner:
Issue for tracking progress on a .proto file for VRS.

larrybabb:
per @reece this should be tracked as a separate project. @ahwagner will ultimately move the branch from here into another repo at that point this will be closed.

ahwagner:
this was moved to a separate repo; see https://github.com/ga4gh/vrs-protobuf

PR #300, opened 2021-05-10
Issue #299: Create a protobuf artefact for VRS

ahwagner:


reece:
@ahwagner @larrybabb Although it's great to have protobuf capability, it increases our maintenance obligation. Before we merge this, let's document our maintenance policy, how we test, etc.

My inclination is to move the y2p code and proto file to a separate repo (e.g., vrs-proto) that depends on vrs but can be versioned and tested independently. That is, make it a particular embodiment of the spec (as with vrs-python), rather than an integral part of the spec.

Issue #258, opened 2021-01-22
Ensure that all classes have tests.


reece:
- [ ] Allele
- [ ] CURIE
- [ ] ChromosomeLocation
- [ ] CopyNumber
- [ ] CytobandInterval
- [ ] DefiniteRange
- [ ] DerivedSequenceExpression
- [ ] Gene
- [ ] Haplotype
- [ ] HumanCytoband
- [ ] IndefiniteRange
- [ ] LiteralSequenceExpression
- [ ] Location
- [ ] Number
- [ ] RepeatedSequenceExpression
- [ ] Sequence
- [ ] SequenceExpression
- [ ] SequenceInterval
- [ ] SequenceLocation
- [ ] SequenceState
- [ ] SimpleInterval
- [ ] Text
- [ ] VariationSet

PR #302, opened 2021-05-13
#301 add constraints

### LINKED ISSUE #301: Constraints on non-negative integer attributes
Currently `CopyNumber`, `RepeatedSequenceExpression`, and `SequenceInterval` all expect non-negative integer values for `copies`, `count`, and `start`/`end`, respectively. However, they use the `Range`, `DefiniteInterval` and `IndefiniteInterval` classes, which are `number` based.

We should add constraints to `CopyNumber`, `RepeatedSequenceExpression`, and `SequenceInterval` to enforce non-negative integers.

ahwagner:
closes #301


reece:
Looks good to me. Still passes vrs-python tests.

For your gray matter sticky notes: When we address #258, we should add tests for constraints too.

reece (approved):
Looks good to me. Still passes vrs-python tests.

For your gray matter sticky notes: When we address #258, we should add tests for constraints too.

PR #321, opened 2021-06-19
Merge changes from 1.2-revisions into v1.2

ahwagner:


schema/y2j
```diff
@@ -6,4 +6,4 @@ import yaml
 
 # I randomly discovered that indent=3, sort_keys=True sorts $schema to
 # the top. I have no idea if that's intentional behavior.
-json.dump(yaml.load(sys.stdin, Loader=yaml.SafeLoader), sys.stdout, indent=3, sort_keys=True)
+json.dump(yaml.load(sys.stdin, Loader=yaml.SafeLoader), sys.stdout, indent=3, sort_keys=False)
```
reece, line 9: Why is `sort_keys=False`? Setting `sort_keys=True` makes diffs much easier.

I think you may have done this because order matters in anyOf/allOf/oneOf blocks and we want to preserve the order that occurs in the yaml. If that's correctly, please add a comment with this rationale.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,29 @@
+name: tests
```
reece, line 1: Thanks for migrating this to github actions.

schema/y2j
ahwagner: I should have clarified that this is a merge PR from `main`, and many of the changes here will appear to be otherwise unrelated and unexplained.

This particular change was addressing #282, based off a documented in-code error.

schema/vrs.yaml
```diff
@@ -22,7 +22,6 @@ definitions:
   # =============================================================================
 
   Variation:
-    additionalProperties: false
```
reece, line 25: Why are you removing additionalProperties?

reece (approved):
Looks good to me.

PR #322, opened 2021-06-19
Issue 270: Align Terminology and Schema

ahwagner:


ahwagner:
@reece and @larrybabb follow this draft PR for major updates to the terminology to synchronize it with the schema. I'm done for tonight; feel free to review and/or jump in with additional commits on branch `270-align-terminology`, I will pick up from this branch tomorrow.

docs/source/terms_and_model.rst
```diff
@@ -1381,23 +1259,20 @@ such as this statement of increased copy number of BRCA1:
   regulatory regions, transcribed regions and/or other functional
   sequence regions.
 
-.. _Quantity:
-
-Quantity
-@@@@@@@@
+Numerics, Comparators, and Ranges
```
larrybabb, line 1262: I prefer that we use the heading **General Purpose Types**. I think we will potentially add to these General Purpose Types over time (or at least until we get through the bulk of the Variation types we need to complete to have a comprehensive spec).  I think "Interval" would also be moved under this heading (and maybe even SequenceExpression.  I think the "identifiable objects" get their own headings for sure as these are more like "Resources" in HL7 FHIR parlance. Why not follow what other standards groups do so as to leverage from what others have learned and gain the benefit of following a plan that we can reference and share.

docs/source/terms_and_model.rst
```diff
@@ -1411,37 +1286,84 @@ Absolute copy number counts may not be smaller than zero.
...
-     - maximum value; inclusive
-
-**Implementation Guidance**
+     - A simple number value representing the range maximum
+   * - type
+     - string
+     - 1..1
+     - MUST be "DefiniteRange"
 
-* If both ``min`` and ``max`` MUST satisfy ``min
```
larrybabb, line 1429: do we need these impl guidance points? Is this a constraint in our schema or is this something we recommend? We should clarify and be consistent as to what level of Impl Guidance we apply across all our types.
```
** Implementation Guidance **
* If the `min` and `max` values are equal then the entire concept will be normalized to a `Number` type with a single `value` equal to the `min` and `max` value. This will assure there is one normalized representation of a single point that may be unintentionally expressed as a DefiniteRange.

File: docs/source/terms_and_model.rst [hunk 3]
```diff
@@ -1411,37 +1286,84 @@ Absolute copy number counts may not be smaller than zero.
...
+     - Limits
+     - Description
+   * - value
+     - number
+     - 1..1
+     - A simple number value representing a range boundary
+   * - comparator
+     - number
+     - 1..1
+     - MUST be one of "<=" or ">=", indicates in which direction
```
larrybabb, line 1361: I think its OK to leave out > and < since the user can simply increase or decrease the value number to make it work. If that's our rationale for leaving out the > and < options then we should probably clarify that in the Impl Guidance section 
```
**Implementation Guidance**
* `>` and `<` options were intentionally omitted. Increasing or decreasing the numeric value to always fulfill the `>=` and `<=` comparators will benefit consistent use and improve normalization.
* The `=` comparator was intentionally omitted. Users should use the `Number` or `DefiniteRange` types if exact values are desired.

File: docs/source/terms_and_model.rst (hunk 2 above)
Comment (larrybabb) [reply docs/source/terms_and_model.rst:1429]: Let me know if you think my Impl Guidance rule above is accurate. It seems to make sense to me that we want our "normalizing routines" to reduce alternative representations down to a common form.

I suppose we could simply fail validation if max <= min. But we may want to allow folks to represent min and max as equal even though they are equivalent to Number. I could go either way I suppose.

File: docs/source/terms_and_model.rst [hunk 4]
```diff
@@ -1575,6 +1497,84 @@ ISCN guidelines [1]_.
...
 @@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
 
+.. _SimpleInterval:
+
+SimpleInterval
+##############
+
+**Computational Definition**
+
+DEPRECATED: A SimpleInterval represents a span of sequence. Positions are
```
larrybabb, line 1507: Question: Will we modify our normalization routine to convert `SimpleInterval`s to `SequenceInterval`s? I suppose that would be a breaking change with what's already released. So sort of a dumb question I guess.
This is an interesting issue which we may want to call out more clearly as it does represent two ways to represent the same intervals.

larrybabb (changes requested):
Looks great. a few minor concerns/recommendations.

docs/source/terms_and_model.rst
```diff
@@ -1169,8 +1044,12 @@ large) reference subsequences specified by a :ref:`SequenceLocation`.
 
 **Computational Definition**
 
-An expression of a sequence that is derived from a referenced
-sequence location.
+An approximate expression of a sequence that is derived from
```
reece, line 1047: Suggested replacement:

A SequenceExpression that is derived from a referenced sequence location. The use of DerivedSequenceExpression suggests that the derived sequence is notional rather than exact, which is typically used for describing large regions for variation concepts where the precision of an exact sequence is unnecessary or undesirable.

docs/source/terms_and_model.rst
reece: WFM

docs/source/terms_and_model.rst
reece: Right. The reason for not having both the inclusive and exclusive forms is that it creates ambiguity. "copy > 2" probably means the same thing as "copy >= 3".

docs/source/terms_and_model.rst
reece: That's an interesting idea, but I'd recommend against that implicit behavior.

Instead, I think we should plan on an explicit conversion process for SimpleInterval → SequenceInterval and SequenceState → LiteralSequenceExpression.

schema/vrs.yaml
```diff
@@ -532,9 +532,10 @@ definitions:
 
   Gene:
     description: >-
-      A reference to an external gene system, used as a location for
-      variation.  Currently, the `ncbigene` namespace is required. See
-      https://registry.identifiers.org/registry/ncbigene.
+      A gene is an authoritative representation of one or more heritable
```
reece, line 535: I prefer the first sentence of the original definition. We don't actually care what the definition is; instead, we care that someone else defined it. I'm ambivalent about the ncbigene part.

reece (changes requested):
I'd like to revert the Gene def

schema/vrs.yaml
larrybabb: If we revert then we should remove the `ncbigene` requirement comment. We could leave it as an example. In reality, we should be pushing for HGNC as that is the standard being promoted by HL7 FHIR guidance even though it too is not strictly required.

docs/source/terms_and_model.rst
larrybabb: I might change the last part "unnecessary or undesireable." to "inconsequential." or maybe "not defining."?

docs/source/terms_and_model.rst
```diff
@@ -400,34 +402,28 @@ referenced, and regardless of order. See
...
+
+*Copy Number Variation* captures the copies of a molecule within a
+genome, and can be used to express concepts such as amplification
+and copy loss. Copy Number Variation has conflated meanings in the
+genomics community, and can mean either (or both) the notion of copy
+number *in a genome* or copy number *on a molecule*. VRS separates
+the concerns of these two types of statements; this concept is a type
+of :ref:`SystemicVariation` and so describes the number of copies in a
+genome. The related :ref:`MolecularVariation` concept can be expressed
+as an :ref:`Allele` with a :ref:`RepeatedSequenceExpression`.
```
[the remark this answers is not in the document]
andreasprlic: Just adding a comment here from somebody who is still learning about vrs.  Repeat expansion variants are different types of variants in my view from CNVs. Repeat expansion variants would mean that a repeated sequence got longer (e.g. FMR1/ fragile X syndrome).  CNVs could be tandem duplications, or interspersed duplications. Using short read sequencing and targeted enrichment, you might not be able to distinguish these two cases, since you might be unlucky and were not able to sequence the breakpoints. One might be just certain e.g. that an exon got duplicated (with uncertain breakpoints somewhere in the introns upstream and downstream).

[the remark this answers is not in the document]
larrybabb: @andreasprlic I'm not sure I understand 100% as I am not a variant scientist or geneticist. We've gone around and around trying to determine the level of specificity we need to add to CNV dupes vs STRs. At this point in the model development we are providing our first version of the CopyNumber class which isn't meant to convey any notion of tandemness. We recognize that folks refer to Dupes (represented with HGVS notation) as a kind of CNV (sometimes, depending on size and possibly assay used to make the call). We struggle with converting HGVS dupe notation to a systemic CNV since it is technically meant to represent the number of copies on a single molecule. 

Our RepeatSequenceExpression is meant to be a tandem repeat and used inside an Allele which is a molecular variation subtype. 

So our struggle is whether it makes sense to represent and transform HGVS dupes into Alleles with RepeatSequenceExpressions thus presuming that HGVS dupes are ALWAYS tandem (when they are not necessarily) versus converting HGVS dupes to CopyNumber systemic variations and making a presumption about the absolute copy count conveyed by an HGVS dupe (problematic on sex chromosomes and not really assured in diploid chromosomes if the zygosity is not explicitly provided).  

The ambiguity that seems natural to experts that refer to CNV dupes is something of a mystery to me still which makes it nearly impossible to specify. 

If you see that the initial forms we are providing now are problematic, let's have a conversation, develop some examples that cover the gambit of possibilities and determine if we can support some or all of them now and provide guidance and wording in the spec to avoid significant difficulties as we evolve the spec.

If you could help by providing those real world examples, and letting us know when you can join a VRS meeting that would be helpful. Also, if you could let us know if you think this is a "show stopper" issue for releasing 1.2 that would be very important to know ASAP.

Thanks.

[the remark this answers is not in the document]
andreasprlic: Hi, Not all CNV dups are in tandem. In fact I recall hearing a story about an example where the difference in knowledge between a tandem DUP vs an interspersed DUP results in a different interpretation (pathogenic vs variant of unknown significance). 

I can try to dig out specific examples for this, if that is helpful. Perhaps a good next step is for me to find some examples and then we can try out how to represent them?

I'd be more than happy to join the next VRS meeting. I am now getting invites from the ga4gh mailer. I assume the invite would come from that?

Thanks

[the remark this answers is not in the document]
larrybabb: yes. we are meeting this Monday at 12-1p ET to review the 1.2 PR and we should have time to discuss this point to make sure we can address any concerns around these issues. It would be ideal if you could attend and bring and example or two. Even if you can't gather the examples, we can discuss the concern properly with the group and determine what course, if any, needs to be taken.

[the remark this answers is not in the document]
reece: "CNV" has many meanings that differ in the size of the repeat unit and what is known about the location of the copies. A key message from the imminent 1.2 release is that we want to distinguish molecular changes, including repeats that we know are in tandem, from those that are systemic, where we know that they exist but not where. So, to be clear, we're not saying that all CNVs are tandem.

In my opinion, the many macroscropic meanings of CNVs -- from nucleotide and exonic tandem repeats to gene-scale duplications anywhere -- make it a difficult term to nail down as a single data structure.

The current direction in VRS is that we really have two kinds of CNVs: RepeatedSequenceExpression, which is a Molecular concept and implies tandemness, and CopyNumber, which is a Systemic concept that makes no assertion about where the copies are located. A CopyNumber even can be inferred from a RepeatedSequenceExpression, but not the converse.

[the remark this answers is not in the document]
andreasprlic: @larrybabb oh interesting. I did not get an invite for the meeting on Monday, but a notification that a meeting last week was canceled. I'd be glad to join, if somebody could forward an invite...

From what @reece says I think this would work. Here a few CNV variants that we could dig out (thanks @jinghuashi ): 

- here an imprecise CNV dup, where we don't know the breakpoints exactly, but an exon got duplicated, most likely in tandem and in-frame `NC_000017.10:g.41209048-?_41209172+?dup`, copy number is 3
Screen Shot 2021-06-24 at 2 10 51 PM

- here another imprecise CNV dup, in tandem, but out of frame `NC_000013.10:g.32944519-?_32945257+?dup` , CN=3
Screen Shot 2021-06-24 at 2 16 48 PM

- An out of frame CNV dup, a precise event, results in a fusion protein `NC_000003.11:g.49161700_49165984dup`, also CN=3
Screen Shot 2021-06-24 at 2 24 39 PM

- For sex chromosomes we prob want to be able to express what is an expected copy number, and what is observed. For a male sample there would be only one copy of X (CNNORM=1), but the pseudo autosomal regions would have CNNORM=2, since they are homologous between X, Y. For a female sample X would have CNNORM=2.

- just to add, here a [reference](https://pubmed.ncbi.nlm.nih.gov/25640679/) about most CNV dups being in tandem, but not all.

As for repeat-expansion variants, for an FMR1 variant (linked to fragile X syndrome), I'd like to be able to say that in a sample we observed e.g. 94 repeats of CGG ([reference](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3151325/)). Something along the lines of `NC_000023.10:g.146993569-146993571[94]`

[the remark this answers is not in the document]
larrybabb: great stuff @reece. stupid question (pardon my ignorance) but is the technology used to make these dup calls assured to know that these copied regions are in phase? If so, then these dups are really a kind of Molecular CNV and not really a Systemic CNV. Should we help the community clearly differentiate between systemic and molecular CNVs? 

Another question ... Are all dups considered a single extra copy or can there be more than one extra copy? 

In clinvar on this variant [GRCh38/hg38 1p34.3(chr1:35934217-36821999)x4](https://www.ncbi.nlm.nih.gov/clinvar/variation/149964/) it seems clear to me that there are 4 copies (somewhere in the genome) of the specified region.  Yet they also represent it down in the HGVS list as NC_000001.11:g.(?_35934217)_(36821999_?)dup.  So is this really a situation where we know there are 2 copies of the region on both instances of the chr1 molecule (in other words a homozygous dupe?). Again, apologies for my lack of knowledge, but getting to the bottom of all of these scenarios and how folks interpret these representations of variants is essential to getting our structures established so everyone can say exactly what they mean to say in a precisely specific or precisely ambigous way - whatever fits the needs of the producer.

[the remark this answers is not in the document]
andreasprlic: @larrybabb I think we generally try to be careful how we report such variants. In some cases we can actually detect if there is an interspersed duplication (which makes them more likely to be benign). In other cases the data suggests that it is likely a tandem duplication, but there is still some small degree of uncertainty so we try to express this using appropriate language. Does that mean it should be a Systemic CNV, rather than a molecular?

I think there can be more than one extra copy, such as your example which could be explained with a potential homozygous dup. Alternatively, a ref-alt heterozygous sample where the alt has 3 copies could also lead to a copy nr. of 4. I am not too familiar with the details of such large events, but I can imagine that there might be a few experimental challenges when trying to resolve the zygosity of such a sample.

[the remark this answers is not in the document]
andreasprlic: Is it possible to represent CNV delins events?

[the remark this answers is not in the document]
reece: Please describe a CNV delins. 

Suffice it to say that there's a lot that VRS can't represent *yet*. Use cases drive development, so let's discuss the use.

reece:
> I think we dropped the MUST fully-justify requirement before; am I misremembering @reece? Dropping this because I think that is the case but can definitely add back in if mistaken.

Yes, we dropped it.

docs/source/terms_and_model.rst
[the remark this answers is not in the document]
reece: Based on yesterday's discussion, I think this is resolved for now.

docs/source/terms_and_model.rst
```diff
@@ -400,34 +402,28 @@ referenced, and regardless of order. See
...
 
 **Computational Definition**
 
-The count of copies of a :ref:`Feature` or
-:ref:`MolecularVariation` subject within a genome.
+The count of discrete copies of a :ref:`Feature`,
+:ref:`MolecularVariation`, or other molecule within a genome.
```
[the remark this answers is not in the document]
reece: I agree with Larry's original comment that "or other molecule" is too vague. Will post change to make more specific.

requirements.txt
```diff
@@ -1 +1,3 @@
-ipython >= 7.5.0
\ No newline at end of file
+pytest
```
reece, line 1: How about listing requirements to .github/workflows/tests.yml?
requirements.txt connotes a python package, which this isn't
I'd prefer to hide this detail from the user.

docs/source/terms_and_model.rst
reece: Renamed to Basic Types.

docs/source/terms_and_model.rst
```diff
@@ -905,128 +886,22 @@ A :ref:`SequenceInterval` with a single start and end coordinate.
...
    * - end
-     - integer
+     - :ref:`Number` | :ref:`IndefiniteRange` | :ref:`DefiniteRange`
      - 1..1
-     - end position
-
-**Implementation Guidance**
+     - Coordinate or coordinate range for the interval end
 
-* Implementations MUST enforce values 0 ≤ start ≤ end. In the case of
```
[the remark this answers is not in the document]
reece: These came back when I merged the 323 branch. I agree that some form of these should be kept. 
Let's keep for now in the SimpleInterval (deprecated) section. We can surgically address this later.

docs/source/terms_and_model.rst
reece: Since this comment is unrelated to aligning terminology and schema, please reraise elsewhere.

docs/source/terms_and_model.rst
reece: Done.

schema/vrs.yaml
reece: Reverted.

requirements.txt
reece: moved to .requirements.txt instead.

reece (approved):

reece:
@larrybabb @ahwagner : I merged examples from 323, resolved conflicts, and addressed all of the comments above.

Please re-review. The goal for this PR was to align terminology and schema. That's done. So, barring significant issues, please approve as-is. We can open a separate round of doc updates if needed.

larrybabb (approved):

larrybabb (approved):

reece:
This seems like a solid step forward. Let's lock this in, and we can iterate as needed.

PR #325, opened 2021-06-23
258 update validation tests

reece:
This branch contains validation tests for all non-abstract classes in VRS.
VRS python HEAD (as of d806dd4, but not earlier) executes these tests as-is.

Tests are framed around VRS classes. In this PR, each class has exactly one test, but having other tests is possible.

For each test, there is an output block that defines the expected output for zero or more functions.

So, for

```
Number:
  -
    in:
      type: Number
      value: 55
    out:
      ga4gh_serialize: '{"type":"Number","value":55}'

```

Interpret as: This test is for the class `Number`. There is one test (`-` is the start of an array). The `in` block defines the class attributes to create an instance. That instance is passed to `ga4gh_serialize`, which is expected to return the given string.

### DIFF
--- a/notebooks/Examples and Validation Tests.ipynb
+++ b/notebooks/Examples and Validation Tests.ipynb
@@ -1,0 +1,1986 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Examples and Validation Tests\n",
+    "\n",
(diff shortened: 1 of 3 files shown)


larrybabb (approved):
Looks good to me. Question for @reece ... Are these examples and validation tests intentionally constrained to testing the canonical json output? Or should these tests include validation for exceptions as well? for example, what if start value is greater then end value? I'm presuming we don't have anything to deal with content validation at that level and we simply rely on the json schema and implementors... Is that right?

reece:
The tests currently cover succuss cases only.

I would like to eventually test failure cases, but don't that in a language neutral way requires more thought.

Issue #341, opened 2021-09-21
Update prefixes table to match ga4gh.yaml


ahwagner:
Currently our prefixes table includes abundance, but not copy number. This needs to be updated.

Issue #340, opened 2021-09-13
DerivedSequenceExpression.reverse_complement fact of fiction?


larrybabb:
[DerivedSequenceExpression](https://github.com/ga4gh/vrs/blob/abd88cd5a99cef598de6f0e4b70b028ebbde2888/docs/source/terms_and_model.rst#derivedsequenceexpression) has a `reverse_complement` attribute in the example but it is not specified in the table containing the definition of the `Information Model`.

Should it be removed from the example or specified in the Information Model?

ahwagner:
Thanks Larry. I think the docs need another round of cleanup to address this and other issues you have raised. Hopefully our work in a single source-of-truth (SOT) document from VRSATILE is translatable here and can help us resolve a lot of this automatically.

Our current SOT should be the JSON Schema, and [we did include](https://github.com/ga4gh/vrs/blob/1.2.0/schema/vrs.json#L666-L668) `reverse_complement` in v1.2.0; so it sounds like this is a documentation issue.

larrybabb:
closing this as no one has raised concerns and as you indicated our VRSATILE work demonstrates the use of `reverse_complement` ideally we would do a patch release of the 1.2 documentation, but at this time 1.2 is obsolete so I don't see the need.

Issue #338, opened 2021-08-11
SequenceInterval does not have an 'Information Model' subsection


larrybabb:
[SequenceInterval](https://vrs.ga4gh.org/en/stable/terms_and_model.html#sequenceinterval) should have an Information Model section similar to CytobandInterval especially since there are several options of classes for the start and end attributes. The inconsistency might confuse new readers of the spec if not leave them with less clarity on the potential options for defining SequenceIntervals.

PR #342, opened 2021-09-22
update y2j

ahwagner:
We have several issues relating to mismatches between the VRS YAML and the information model, e.g. #338, #340, #341.

To address this, I would like to bring over translation tooling from VRSATILE that takes an authoritative YAML using an extended syntax to generate both JSON Schema and restructured text artifacts. On the current YAML using a strict JSON Schema syntax, this has no functional impact on the generated JSON Schema, making it fully compatible with the original `y2j` script.

This PR will provide the foundation for extending VRS to use the extended syntax and automatic documentation generation.


larrybabb (approved):
great job, alex.

PR #344, opened 2021-09-23
remove reference to develop branch

ahwagner:
The develop branch for community review is an outdated process that we no longer follow. Recommend removing since the develop branch no longer exists.


larrybabb (approved):
do it

PR #352, opened 2021-10-05
Issue 350

### LINKED ISSUE #350: Create upstream pre-processed authoritative source
Revert vrs.yaml to be a yaml representation of vrs.json, and add upstream authoritative source for y2j.py

ahwagner:
closes #350. Also demonstrates the utility of the new tooling; accidentally put `id` as required in an earlier commit; correcting to `_id` automatically updates corresponding information models and JSON Schema!


ahwagner:
Closed & reopened this PR to kick off RTD preview auto-build.

schema/Makefile
```diff
@@ -5,10 +5,13 @@
 YAMLS:=vrs.yaml
 JSONS:=${YAMLS:.yaml=.json}
 
-all: ${JSONS} defs
+all: vrs.json defs
```
reece, line 8: What about ga4gh.json?

schema/defs/Variation.rst
```diff
@@ -16,9 +16,9 @@ A representation of the state of one or more biomolecules.
       - Description
    *  - _id
       - :ref:`CURIE`
-      - 0..1
+      - 1..1
```
reece, line 19: `_id` is optional. Should be `0..1`, right?

reece (approved):
Looks good. Thanks for the quick action!

larrybabb:
> Looks good. Thanks for the quick action!

That's what they said?

larrybabb (approved):

schema/Makefile
ahwagner: AFAIK, ga4gh.json isn't used anywhere, and unlike vrs.json, isn't JSON Schema.

It's not a big deal to put it back in, but is it actually our intent to provide this file for some purpose, or is the YAML sufficient?

schema/defs/Variation.rst
ahwagner: Oh wow. Thanks @reece, great catch. Will fix.

schema/Makefile
reece: I provided for completeness (json and yaml pairs), but your logic is sound and that file should probably move to ga4gh core anyway. So, let's keep as-proposed and we'll revive if needed.

PR #354, opened 2021-10-30
Use shared metaschema tooling

ahwagner:
Currently we maintain a version of the translation tooling in VRS, and another in VRSATILE. This PR points at tooling residing in VRSATILE, leaving this repo to focus on the specification itself.


larrybabb (approved):
looks good to me.

schema/vrs.yaml
```diff
@@ -64,17 +64,18 @@ definitions:
...
   Haplotype:
-    description: A set of non-overlapping Alleles that co-occur on the same molecule.
+    description: A set of non-overlapping Allele members that co-occur on the same
+      molecule.
     additionalProperties: false
     type: object
     properties:
-      _id:
-        $ref: '#/definitions/CURIE'
+      _id: *id001
```
reece, line 72: Why use refs only here? We should use consistently.

reece (changes requested):
This PR makes VRS dependent on a ancillary part of VRSATILE, which is an extension of VRS (kinda). That seems circular.

Many changes to VRSATILE will be completely irrelevant to VRS, but will now need to be considered as having possible implications to the VRS schema. For example, a change in bowels of vrules might incorrectly imply a change in VRS through the use of vrsatile tools. That will exacerbate maintaining VRS and have  ripple consequences for vrs-python.

A much better structure is to create a separate repo, e.g., vr-tools, that is used by both. What do you think of that?

While we're at it, publishing vr-tools to pypi would allow us to treat it as a conventional Python dependency rather that using submodules.

schema/vrs.yaml
ahwagner: This is only done here because I've only added Haplotype and Allele to inherit from Variation so far. As I add others, inherited properties (really, just `_id`) will be used by ref in all applicable classes.

Of course, this is just a YAML convention and in JSON Schema, each instance will remain fully described.

ahwagner:
These commits add the meta-schema processing scripts and module through PyPI, removing any circularity from sub-repositories.

As [mentioned above](https://github.com/ga4gh/vrs/pull/354#discussion_r740324031), additional refs will be used as the meta-schema is extended to cover other variation forms beyond Allele and Haplotype. We will eventually reach consistency. However, to help separate concerns of the tooling from the schema itself, I would like to keep this PR focused on the adoption of the VRSATILE meta-schema processing tooling to replace the in-kind version currently in the VRS schemas directory.

reece (changes requested):
This still feels circular to me. The fact that the vrsatile tools is a package doesn't really change that vrs depends on (the tools of) vrsatile, and that (the spec of) vrsatile depends on vrs.  I'll pick this up in slack.

reece (approved):
Looks good. Thanks for the changes Alex.

Issue #355, opened 2021-11-17
species vs species_id in ChromosomeLocation


korikuzma:
In [terms_and_model.rst](https://github.com/ga4gh/vrs/blob/main/docs/source/terms_and_model.rst), the field is `species`. In the [vrs schema](https://github.com/ga4gh/vrs/blob/main/schema/vrs.yaml#L217), the field is `species_id`. @ahwagner Is the field actually `species_id` and it's just a typo in the docs?

ahwagner:
The field is indeed supposed to be `species_id`; just another reason why we are transitioning over to the meta-schema / single definition tooling for both docs and schema.

Thank you for pointing this out; as it so happens I have been working on the VRS meta-schema transition today and will do `Locations` next to cover this.

korikuzma:
This is stale. We can close.

PR #356, opened 2021-11-21
Transition docs to metaschema

ahwagner:
This PR addresses inconsistencies between the docs page and the JSON Schema by relying on a single source-of-truth, enabled by our meta-schema tooling.


ahwagner:
f1e4ce939202346638a479c86dc6a09edadfec33 addresses #355

ahwagner:
f1bf3f5 addresses #338

ahwagner:
bb11ae1 addresses #340

schema/vrs-source.yaml
```diff
@@ -184,26 +184,31 @@ definitions:
     additionalProperties: false
     type: object
     description: >-
-      The count of discrete copies of a Feature, Molecular Variation, or
-      other molecule within a genome.
+      The count of discrete copies of a subject :ref:`MolecularVariation`,
```
larrybabb, line 187: Should this be re-written as...

> The 'absolute' count of discrete copies of a MolecularVariation, Feature, SequenceExpression or CURIE reference contained within a system (i.e. genome, cell, etc...).

I'm not sure the current phrasing is clear. What does 'to any of these' mean in the context of count of discrete copies of a subject...?

schema/vrs-source.yaml
```diff
@@ -184,26 +184,31 @@ definitions:
...
           - $ref: "#/definitions/CURIE"
+        description: >-
+          Subject of the Copy Number object
       copies:
         oneOf:
           - $ref: "#/definitions/Number"
           - $ref: "#/definitions/IndefiniteRange"
           - $ref: "#/definitions/DefiniteRange"
+        description: >-
+          The integral number of copies of the subject in a genome
```
larrybabb, line 211: should be ..."copies of the subject in a system (ie. genome, cell, ...)"

schema/vrs-source.yaml
```diff
@@ -304,34 +331,39 @@ definitions:
...
   SequenceInterval:
     description: >-
-      A SequenceInterval represents a span of sequence. Positions are
-      always represented by contiguous spans using interbase coordinates.
-
-      SequenceInterval is intended to be compatible with that in Sequence Ontology
-      (`SO:0000001<http://www.sequenceontology.org/browser/current_svn/term/SO:0000001>`_),
-      with the exception that the GA4GH VRS SequenceInterval may be zero-width. The SO
-      definition is for an "extent greater than zero".
+      A SequenceInterval represents a span of :ref:`Sequence`. Positions are
```
larrybabb, line 338: it's not the "span of a Sequence". it's the "span on a Sequence".  We should clarify that the Interval classes are not "sequences" or "segments/spans of sequence" they are merely the interval or span that is used by SequenceLocation and CytobandLocation to specify the start/end of the segment that is being located.

This is the set of classes that we want to roll up into sequenceLocation and cytobandLocation so we can avoid this unnecessary nesting and confusion.

DerivedSequenceExpression is the "span of a sequence"

schema/vrs-source.yaml
```diff
@@ -411,18 +443,27 @@ definitions:
 
   CytobandInterval:
     description: >-
-      A contiguous region specified by chromosomal bands features.
+      A contiguous region specified by cytobands. The region includes
```
larrybabb, line 446: For consistency this should be stated similarly to SequenceLocation "the span on a chromosome bounded by ctyoband endpoints" or something to that affect. It is not a region or a cytoband itself. It is an "interval" that is strictly used by the ChromosomeLocation to define a location on a chromosome based on cytoband nomenclature.

schema/vrs-source.yaml
```diff
@@ -613,90 +663,109 @@ definitions:
...
         type: number
+        description: The minimum value; inclusive
       max:
         type: number
+        description: The maximum value; inclusive
     required: ["type", "min", "max"]
 
+  IndefiniteRange:
+    description: >-
+      A half-bounded range of numbers represented as a number bound and
```
larrybabb, line 677: While I agree that limiting this to "<=" and ">=" is correct for Integer values. It may not be sufficient for Real numbers. When Integers are the class of Numbers in the value field (or min and max fields for DefiniteRange) the equality portion of the comparator can be made on a precise Integer. But with Real numbers if one wanted to convey the concept of "less than 4" they wouldn't be able to express that precisely enough using the equality symbol (it could be 3.99999 or 3.99999999 or ...).
So I wonder if we jumped the gun on this or if we need to re-litigate this?
If we add back the ">" and "<" as options then we will introduce dual ways to convey the same Integer Ranges by either using the precise endpoint with the gt/lt equal symbol comparators or the prev/next endpoint with the gt/lt symbols. Ugh.

larrybabb (approved):
@ahwagner Please feel free to merge. I made a few comments, not so much about anything new, more about things I probably should have caught in the past. We can discuss as we move forward.
( i  do think the SequenceInterval and CytobandInterval definitions should be tweaked as I stated in my comments )

schema/defs/vrs/ChromosomeLocation.rst
```diff
@@ -0,0 +1,38 @@
...
+      - 1..1
+      - MUST be "ChromosomeLocation"
+   *  - species_id
+      - :ref:`CURIE`
+      - 1..1
+      - :ref:`CURIE` representing a species from the `NCBI species taxonomy <https://registry.identifiers.org/registry/taxonomy>`_. Default: "taxonomy:9606" (human)
+   *  - chr
+      - string
+      - 1..1
+      - The symbolic chromosome name. For humans, For humans, chromosome names MUST be one of 1..22, X, Y (case-sensitive)
```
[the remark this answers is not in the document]
ahwagner: Mitochondria are not karyotyped alongside the nuclear chromosomes in cytogenomic assays and so they are not part of the specification.

schema/vrs-source.yaml
ahwagner: Added #362 to address later.

schema/vrs-source.yaml
ahwagner: Added #361 to address later.

schema/vrs-source.yaml
ahwagner: addressed in e2447ba

Issue #377, opened 2022-02-01
Need a write up on why we don't VOCA normalize "same as reference" alleles.


larrybabb:
@reece Would you have notes or be able to draft a paragraph or two on the policy decision we made regarding the rationale not to VOCA normalize "same as reference" variation? Our documentation could use this. Some folks have been asking.

Currently if someone tries to VRSify an allele that has a change from the reference we do a VOCA normalization to derive the fully normalized span of reference sequence and the dervied variation that would produce the change originally sought.

However, if someone tries to VRSify an allele that is the same as reference we do not do any VOCA normalization (but we do verify the sequence matches if provided...i assume) and simply use the exact span with the ref sequence as the state.  If this span was within a repeated region it would NOT be expanded fully like a change from reference would.

larrybabb:
this was completed and merged in PR #381

PR #381, opened 2022-03-06
explain ref agree normalization rules

reece:
Finally got around to Larry's requests for explaining ref agree normalization rules.

Instead of writing a new section, I overhauled an existing related section (`should-normalize`).

### DIFF
--- a/docs/source/appendices/design_decisions.rst
+++ b/docs/source/appendices/design_decisions.rst
@@ -52,23 +52,46 @@ Implementations should normalize Alleles
 
 VRS STRONGLY RECOMMENDS that Alleles be :ref:`normalized
 <normalization>` when generating :ref:`computed identifiers
-<computed-identifiers>` unless there is compelling reason to do otherwise.
-Those reasons are the subject of this section.
+<computed-identifiers>` unless there is compelling reason to do
+otherwise.  Those reasons are the subject of this section.
 
-:ref:`Fully-justified Normalization <normalization>` is the process of comparing a span of reference sequence to a sequence state (often the alternative sequence).  Normalization consists of two steps: trimming and shuffling.  In the trimming step, common flanking prefix and suffix sequences are removed.  For example, a CAG-to-CTG Allele would be trimmed to merely A-to-T, with the position adjusted accordingly.  There are four cases of the resulting sequences:
+:ref:`Fully-justified Normalization <normalization>` is the process of
+comparing a span of reference sequence to a sequence state (often the
+alternative sequence).  Normalization consists of two steps: trimming
+and shuffling.  In the trimming step, common flanking prefix and
+suffix sequences are removed.  For example, a CAG-to-CTG Allele would
+be trimmed to merely A-to-T, with the position adjusted accordingly.
+There are four cases of the resulting sequences:
 
-  1. The trimmed sequences are empty: The Allele refers to reference state.
-  2. The trimmed sequences are non-empty: The Allele is a substitution (perhaps multi-residue).
+  1. The trimmed sequences are empty: The Allele refers to reference
+     state.
+  2. The trimmed sequences are non-empty: The Allele is a substitution
+     (perhaps multi-residue).
   3. The reference sequence is empty: The Allele is a net insertion.
   4. The state sequence is empty: The Allele is a net deletion.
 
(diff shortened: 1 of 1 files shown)


larrybabb (approved):

reece:
Did y'all (i.e., @ahwagner , @andreasprlic ) see that I addressed #377 in this PR a few weeks ago?
@larrybabb approved

docs/source/appendices/design_decisions.rst
```diff
@@ -32,10 +32,10 @@ Allele Rather than Variant
 The most primitive sequence assertion in VRS is the :ref:`Allele`
 entity. Colloquially, the words "allele" and "variant" have similar
 meanings and they are often used interchangeably. However, the VR
-contributors believe that it is essential to distinguish the state of
-the sequence from the change between states of a sequence. It is
+contributors believe that it is essential to distinguish the *state* of
+the sequence from the *change between states* of a sequence. It is
```
ahwagner, line 36:
```suggestion
contributors assert that it is essential to distinguish between the *state of*
a reference sequence from the *change from* a reference sequence. It is
```

docs/source/appendices/design_decisions.rst
```diff
@@ -32,10 +32,10 @@ Allele Rather than Variant
...
 imperative that precise terms are used when modelling data. Therefore,
-within VRS, Allele refers to a state and "variant" refers to the change
+within VRS, "Allele" refers to a state and "variant" refers to the change
```
ahwagner, line 38:
```suggestion
within VRS, "allele" refers to a state of a reference sequence and "variant" refers to a change
```

docs/source/appendices/design_decisions.rst
```diff
@@ -32,10 +32,10 @@ Allele Rather than Variant
...
 from one Allele to another.
```
ahwagner, line 39:
```suggestion
from a reference sequence.
```

docs/source/appendices/design_decisions.rst
```diff
@@ -113,6 +74,55 @@ occurs in a low-complexity region, but rather describes the final and
...
+
+Implementations should normalize Alleles
+@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@
+
+VRS STRONGLY RECOMMENDS that Alleles be :ref:`normalized
+<normalization>` when generating :ref:`computed identifiers
+<computed-identifiers>` unless there is compelling reason to do
+otherwise.  Those reasons are the subject of this section.
+
+:ref:`Fully-justified Normalization <normalization>` is the process of
```
ahwagner, line 87:
```suggestion
:ref:`Allele Normalization <normalization>` is the process of
```

docs/source/appendices/design_decisions.rst
```diff
@@ -113,6 +74,55 @@ occurs in a low-complexity region, but rather describes the final and
...
+comparing a span of reference sequence to a sequence state (often the
+alternative sequence).  Normalization consists of two steps: trimming
```
ahwagner, line 89:
```suggestion
alternative sequence) and resolving that span to an unambiguous form.  The fully-justified Allele normalization in VRS consists of two steps: trimming
```

ahwagner (changes requested):
Thanks for the ping on this thread @reece, I missed the original.

I think these changes accurately represent prior discussion on these points. I do think that the distinction between Allele and variant is artificial, particularly as we do not normalize reference-agree and reference-disagree Alleles the same. In addition, we (for well-justified reasons) represent Alleles as a state on a reference sequence, inherently comparing the state to some underlying reference.

To be very clear about what I mean, for two (short) reference sequences:

Reference 1: AACTA
Reference 2: AACGA

A resultant sequence of AACCA may be expressed as two different Alleles, conceptually:
- A `C` at interresidue coordinates `<3,4>` on `Reference 1`
- A `C` at interresidue coordinates `<3,4>` on `Reference 2`

Each of these result in different digests, as the information of the reference sequence is inherent to the representation of each Allele. The primary distinction between Allele and similar entities from other major variant formats is that we:
1. Do not represent the reference state as it is retrievable
2. Do not provide special character representation (e.g. `=`) of the reference-match state

I am not advocating that we compute resultant sequence to create digests; the process is far too cumbersome to be practical in implementation and the gains would be nearly non-existent (i.e. analogous reference sequences are too dissimilar for the above scenario to realistically occur). But I think we should revise the text here to better capture the notion of allele as a reference-match agnostic state instead of a reference-sequence agnostic state. The edits I have suggested here reflect this subtlety.

reece:
Looks good to me (but I'm not a reviewer and am totally powerless! Yay!)

ahwagner (approved):
Changed applied per 👍 from @reece

Issue #277, opened 2021-02-22
RelativeCopyNumber


ahwagner:
Revisit Relative CopyNumber statements, as seen in cytogenetics resources and WRT X-chromosome abnormalities.

ahwagner:
@mcannon068nw follow this thread.

mbaudis:
+1

mbaudis:
As mentioned in an exchange w/ @ahwagner the **representation of relative copy number states is essential** since:

* most analyses do not deliver copy number counts - in fact the use of "copies"  is extremely unusual
* CN estimates will be influenced by clonality, impurity and ploidy
* even correct CN counts can frequently (cancer genomes...) only be interpreted by knowing the base ploidy level

A main logical paradigm of CN analyses and representation is the _relative_ level with respect to a baseline (i.e. ploidy level). In cancer genomics there is pretty much consistent use of a limited set of CN levels:

* homozygous deletion (i.e. estimated 0 copies in any ploidy)
* deletion (i.e. fewer copies than the assumed baseline; `CN=1` for diploid)
* duplication / low level gain (i.e. one or few more copies than baseline; operationally one can assume this as "not more than a duplication of the base CN count", i.e. `4` in a diploid genome)
* amplification / high level gain : from ~5 to any number - possibly in the hundreds - of copies of a genomic region. However, there is no clear definition of an `amplification` threshold and some definitions may include the regionality and exclude e.g. events leading to multiple copies of a chromosome

In practice, the current lack of a relative indication of CN state prohibits the use of the schema for most real world applications representing CNV events (or require to use fake values).

#### Changes needed

1. a class describing the relative copy number state (REQUIRED)
2. a representation of the base copy number at the location in the given sample (OPTIONAL; e.g. 2 for autosomes in diploid cells, 1 for X/Y in males, 3 for a triploid cell line with e.g. 69,XXX etc.)

The best option would be to have an ontology for such classes & SO should be the place? However, CNV representation there is confusing & incomplete.

#### Minimal pseudo-ontology for CNVs

```
id: CNVO:000001
label: copy number assessment
  id: CNVO:000002
  label: base ploidy
   id: CNVO:000004
   label: copy-neutral loss of heterozygosity
  id: CNVO:000003
  label: copy number variation
    id: CNVO:000005
    label: copy number loss
      id: CNVO:000007
      label: low-level copy number loss
      id: CNVO:000008
      label: homozygous deletion
    id: CNVO:000006
    label: copy number gain
[...]

PR #382, opened 2022-03-08
relative copy number

ahwagner:
draft of the relative copy number class


korikuzma (approved):

jsstevenson (approved):

larrybabb (approved):
my only suggestion is that we provide better examples, but we really need to do this throughout the spec. so i'll make a separate ticket that we can address later.

ahwagner:
closes #277

PR #387, opened 2022-03-24
386: revise base64url description

ahwagner:


jmarshall:
:shrug: I think it's good that the introductory paragraph uses familiar terms, while steps 1–3 are more precise.

So I would just delete /URL/ leaving `and [Base64](link) encoding`. Or write out `and the URL-friendly variant of [Base64](link) encoding` if you wanted to avoid accusations of conceivably being misleading.

But it's all a bit bikesheddy, so whatever is good.

jmarshall:
Yeesh, how on earth did the readthedocs run for the first commit succeed? :smile: Has there been a massive readthedocs update in the last fourteen hours?

ahwagner:
I think `Jinja2` is the culprit, made a breaking change in a minor version release an hour ago that Sphinx wasn't expecting: https://pypi.org/project/Jinja2/#history

jmarshall:
Yep. Looks like Sphinx 4.x would be insulated against the change BTW: sphinx-doc/sphinx@e5c905c9617d3e10d6b37d6548761c9f1ceb48b9.

ahwagner:
@larrybabb and/or @andreasprlic you good with these changes?

larrybabb (approved):
sorry for the delay.

PR #394, opened 2022-07-11
VRS Hackathon Genotype draft

ahwagner:
This is a draft model of the Genotype class that was refined and tested during the ISMB 2022 VRS Hackathon.

This draft addresses #393.

TODO: add maturity labels to classes.


schema/vrs.json
```diff
@@ -119,11 +123,12 @@
             "type": {
                "type": "string",
                "const": "Haplotype",
+               "default": "Haplotype",
                "description": "MUST be \"Haplotype\""
             },
             "members": {
                "type": "array",
-               "minItems": 1,
+               "minItems": 2,
```
mbaudis, line 131: Probably has been discussed - but what about haploid genomes/chromosomes?

mbaudis (approved):
Apart from my question regarding the min 2 alleles for a haplotype - great :-)

schema/defs/vrs/GenotypeMember.rst
```diff
@@ -0,0 +1,28 @@
+**Computational Definition**
+
+A class describing a :ref:`Genotype` `member`.
```
larrybabb, line 3: More specific description would be helpful to readers I think.  Something like "A representation of the number of copies of a given  haplotype or allele that make up the members of a genotype. This structure type is only used within the Genotype data type."

We should also provide 2 examples of this so that we can emphasize how the copies is used to represent Het vs Hom alleles or haplotypes.  It will also be useful to reuse the examples in the Genotype examples so that we can clearly show the purpose of the Genotype.copies from the GenotypeMember.copies which will most certainly be a point of confusion.

We may need implementation guidance as well, but it may make sense to simply have all the implementation guidance in the Genotype section.

schema/defs/vrs/Haplotype.rst
```diff
@@ -26,5 +26,5 @@ Some Haplotype attributes are inherited from :ref:`Variation`.
       - MUST be "Haplotype"
    *  - members
       - :ref:`Allele` | :ref:`CURIE`
-      - 1..m
+      - 2..m
```
larrybabb, line 29: @ahwagner is this a breaking change? Or are we not worrying about any breaking changes until we establish the maturity model? 

We may want to call out that this does invalidate previous versions that had haplotypes of 1 implemented.  I'm not sure how to do this, but we should consider annotating these types of "minor" changes that are borderline or actual breaking changes that we are including because we feel as if the model is in an Draft or Trial Use state.

schema/ga4gh.yaml
```diff
@@ -23,7 +23,7 @@ identifiers:
     Allele: VA
     VariationSet: VS
     Text: VT
-    # Genotype: VG
+    Genotype: GT
     Haplotype: VH
```
larrybabb, line 27: since we are diverging from the Vx form of prefixes for Genotype, we may want to change Haplotype to HT? Again, we can put this under the heading of "improvement to draft data types". It would provide some consistency and justify the "GT" prefix.  Or maybe we simply call all of this "Trial Use" and fix it in 2.0?

schema/ga4gh.yaml
```diff
@@ -23,7 +23,7 @@ identifiers:
     Allele: VA
     VariationSet: VS
     Text: VT
```
larrybabb, line 25: @ahwagner Here's a whopper.  I think we should consider "Text" or "VariationText" to be a universal and ambigous thing that can be used for any "text string" not just for variation. This would allow us to using it for "Conditions", "Drugs" and other value-object-like domain entities that we will be using.  Just because it's under the VRS umbrella does not mean it is specialized to be a Variation (I think).  Unless we think that subclassing these utilitye "text" concepts to domain specific types (i.e. ConditionText, DrugText, etc...) is important for folks to know that it meant to be a Condition, Drug, etc... .  Just thinking that if we wanted to do this then we would change the prefix to "TX" or "TXT" and modify the documentation to emphasize that it is truly a utility class that open to interpretation and will require the producer and consumer to "imply and infer" it's meaning (which is what they do anyway).

larrybabb: this is most definitely a "Draft" data type currently, but is something I will be implementing in ClinVar for Variation and Conditions.

schema/vrs-source.yaml
```diff
@@ -240,6 +246,43 @@ definitions:
 
     required: [ "subject", "copies" ]
 
+  Genotype:
+    description: >-
```
larrybabb, line 250: For just about all of our "Variation" data types we have a "Description" that shows right under the datatype main heading and a "Computational Definition". It seems that our Genotype is missing the former "general" description and this description is showing up as the "Computational Definition". For quality and consistency purposes we should follow the approach in "[Haplotype](https://vr-spec--394.org.readthedocs.build/en/394/terms_and_model.html#haplotype)" shown here

> Haplotypes are a specific combination of Alleles that are in-cis: occurring on the same physical molecule. Haplotypes are commonly described with respect to locations on a gene, a set of nearby genes, or other physically proximal genetic markers that tend to be transmitted together.
> 
> Computational Definition
> 
> A set of non-overlapping [Allele](https://vr-spec--394.org.readthedocs.build/en/394/terms_and_model.html#allele) members that co-occur on the same molecule.

schema/vrs-source.yaml
```diff
@@ -240,6 +246,43 @@ definitions:
...
+        const: "Genotype"
+        default: "Genotype"
+        description: >-
+          MUST be "Genotype"
+      members:
+        type: array
+        uniqueItems: true
+        minItems: 1
+        items:
+          $ref: "#/definitions/GenotypeMember"
```
larrybabb, line 267: this is the reference that is not showing as a link (I think).

schema/vrs-source.yaml
```diff
@@ -240,6 +246,43 @@ definitions:
...
+        description: >-
+          Each GenotypeMember in `members` describes a :ref:`MolecularVariation`
```
larrybabb, line 269: This is a slightly better definition than the one that is currently associated with the GenotypeMember section.

schema/vrs-source.yaml
```diff
@@ -240,6 +246,43 @@ definitions:
...
+          and the count of that variation at the locus.
+      copies:
+        oneOf:
+          - $ref: "#/definitions/Number"
+          - $ref: "#/definitions/IndefiniteRange"
+          - $ref: "#/definitions/DefiniteRange"
+        description: >-
+          The total number of copies of all :ref:`MolecularVariation` at this locus,
```
larrybabb, line 277: Is this an accurate description of the copies field? Isn't the Genotype.copies really about the "ploidy" of the system? By saying it is the total number of copies of all MolecularVariation at this locus it presumes a couple things: 1) all GenotypeMembers occupy the same locus (which we are not requiring - I think), and 2) the total number of copies of all MolecularVariation might be misunderstood to not mean "the number of instances of the molecules that contain any one of the GenotypeMembers which make each copy of the members in-trans.".  
It seems like we need to be clearer about what's going on here. I know we get (I think), but we need to explain this very clearly in text so as to reduce confusion.


We should probably add Implementation Guidance to say things like "must be greater than or equal to 1" and "the sum of the member's copies should always be less than or equal to the genotype.copies" and "if the genotype.copies is greater than the sum of the member's copies at any given locus then it can be inferred that the state at the non-specified copies is not explicitly represented."

larrybabb (commented):
I think these changes should be made before we approve a "1.3" release, but I'm unclear if we are just trying to get this into the main branch for now with the intention of doing a 1.3 later.  If this is about approving 1.3 then I would say we need to make some of these changes. Which we should do together as they are a bit subjective and should include some discussion to finalize.

schema/defs/vrs/Genotype.rst
```diff
@@ -0,0 +1,34 @@
...
+      - 0..1
+      - Variation Id. MUST be unique within document.
+   *  - type
+      - string
+      - 1..1
+      - MUST be "Genotype"
+   *  - members
+      - :ref:`GenotypeMember`
+      - 1..m
+      - Each GenotypeMember in `members` describes a :ref:`MolecularVariation` and the count of that variation at the locus.
```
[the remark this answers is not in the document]
larrybabb: The GenotypeMember.copies attribute is a very important element in that it allows us to specify the copies of a given Allele or Haplotype versus having to re-specify the same Allele or Haplotype 2 or more times, including the ability to use ranges to say that you know there are more than a certain number of in-trans copies, etc... We discussed this at the hackathon and concluded that this GenotypeMember sub-structure was more compact. 

If you think this is a suboptimal conclusion, please raise more specific concerns (we can also discuss at our next meeting of course).

schema/vrs.json
larrybabb: This `Haplotype` data type is a molecular variation of in-cis members that are comprised of Alleles. So it should work for any system haploid, diploid, etc... We made this specific change to avoid the confusion of treating a `Haplotype` of 1 member `Allele` to be the same thing as a single `Allele`.  Since both `Allele` and `Haplotype` are subtypes of `MolecularVariation` you should always be able to use `Allele` where you would otherwise use a `Haplotype` of 1 member, thus we made haplotype 2...m to avoid the potential for dual representation of a single allele.

schema/defs/vrs/Genotype.rst
```diff
@@ -0,0 +1,34 @@
...
+   *  - copies
```
reece, line 31: Are two Genotypes the same if their `members` are identical but their `copies` are not?

My recommendation would be to drop the copies for now unless you know that you have a use for them.

reece (approved):

schema/defs/vrs/Haplotype.rst
ahwagner: This doesn't change digests of valid Haplotypes–only reduces the number of Haplotypes that are considered valid. I see this as the sort of thing that can be easily remedied in software–Haplotypes with 1 member will just be transformed into the member Allele.

We are avoiding any changes in 1.x that alter digests (though we have deprecated / replaced Allele properties already with SequenceExpressions), but I think anything else is fair game.

schema/defs/vrs/Haplotype.rst
```diff
@@ -26,5 +26,5 @@ Some Haplotype attributes are inherited from :ref:`Variation`.
       - MUST be "Haplotype"
```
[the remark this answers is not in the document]
ahwagner: I am not sure I understand this concern. Why would `Haplotype.type` be anything other than `"Haplotype"`?

schema/defs/vrs/GenotypeMember.rst
```diff
@@ -0,0 +1,28 @@
...
+      - 0..1
+      - MUST be "GenotypeMember".
+   *  - copies
+      - :ref:`Number` | :ref:`IndefiniteRange` | :ref:`DefiniteRange`
+      - 1..1
+      - The number of copies of the `variation` at a :ref:`Genotype` locus.
+   *  - variation
+      - :ref:`Allele` | :ref:`Haplotype`
+      - 1..1
+      - A :ref:`MolecularVariation` at a :ref:`Genotype` locus.
```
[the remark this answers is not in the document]
andreasprlic: Same here, couldn't this just require a MolecularVariation?

ahwagner:
In addition to the above comments (thank you all!) I have two additional issues with this model, and how it would be used in VRS, that need to be addressed before I think we can merge this.

The first is an issue with `GenotypeMember`. I like this as a [Basic Type](https://vr-spec--394.org.readthedocs.build/en/394/terms_and_model.html#basic-types), and I think it would be a mistake to make this identifiable as a dependent structure for representing the number of Alleles/Haplotypes. However, leaving it as non-identifiable (as it currently is) means that the `Genotype.members` array will not be sorted during serialization (but it should be). To address this, we must do one of the following:

1. consider alternative structures for conveying the number/range of a given Allele / Haplotype in a Genotype
    a. possibly use the CopyNumber class instead of GenotypeMember to assign the copies/count to the MolecularVariation? I think this isn't correct and doing so would conflict with the intended purpose of a Genotype, but it does solve the above technical issue cleanly.
    b. repeat the MolecularVariation explicitly inside the Genotype (doing so would remove our ability to represent definite/indefinite range counts of a given Allele / Haplotype within the Genotype)
2. revise our rules / add custom attributes to JSON Schema array properties indicating whether or not to preserve the order of ANY array during serialization (generalizable strategy that can help elsewhere – my preference)
3. make GenotypeMember identifiable (ick)
    a. or make it sortable via the nested GenotypeMember.variation digest (complex)

My second issue is with the semantics and implementation guidance for this data class. We have described GenotypeMembers as "_in-trans_" but I think we need to be very clear about what this means. Revisiting the point @rrfreimuth raised, most modern assays allow us to estimate the number of any allele that _aligns to_ a reference locus, even if the origin of one or more copies of that allele is at a duplicated locus elsewhere. In cancer, we often see this at high copies, but in my experience we also don't care about the semantics of "_in-trans_" / "phasing" and the notion of "Genotype" when evaluating such loci–just the total number of copies of that Allele system-wide.

So I think we should add some implementation guidance to accommodate this, clarifying that all Alleles / Haplotypes represented in a Genotype are presumed to be _in-trans_ on homologous sequences, and _therefore are constrained to representation on the same sequence_. We should also add to our implementation guidance that no two GenotypeMember objects can have the same Allele subject (unless we change the model to accommodate my first concern above).

schema/vrs.json
larrybabb: @mbaudis I just realized that I may have misunderstood your question.  To be clear, the `minItems` constraint is about the the cardinality of _in-cis_ allele members for the Haplotype being defined. It is not a ploidy count or anything like that. Haplotypes by definition are haploid.  We are using the Genotype class to deal with genome ploidy representation. Refer to the `Genotype.copies` for this attribute (btw - we are considering changing the attribute name from copies to something that more clearly represents the number of _in-trans_ occurrences of each`Gentotype.member`.  Let us know if this resolves your issue. If not, please restate so we can be sure to understand any additional concerns.

schema/defs/vrs/Genotype.rst
larrybabb: No. If the copies are different in any way then they are not precisely the same genotypes. If we are trying to acheive the level of exactness as we do for Alleles then we are asserting that the _ploidy_ or `copies` of the _in-trans_ instances of the underlying loci that covers the region containing the members is not the same across two different Genotypes (even if the members are identical) then it is not considered an exact match.

We believe this adheres to the VRS principles of precision.

schema/vrs-source.yaml
larrybabb: @ahwagner we need to add additional implementation guidance and a reference to how these GenotypeMembers are first _Normalized_ ahead of being digested.  For example, if someone creates a Genotype with 2 separate GenotypeMembers that have the same exact Allele, then it should be reduced into a single GenotypeMember with a count that is the sum of the individual member's counts (which could involve combining indefinite ranges, definite ranges and numbers, of course).  

I do think each Variation class should have a section related to **computed digest notes** which calls out any caveats to be resolved in _normalizing_ the representation before it is digested.  I think as the model gets more and more complex these _normalization_ notes will be critical for generating matching digests for situations that the schema validation cannot catch but are still truly equivalent.

schema/defs/vrs/GenotypeMember.rst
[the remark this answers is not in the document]
larrybabb: We decided that it is not pragmatic to force every copy of the same MolecularVariation that is _in-trans_ to be represented explicitly. We also recognize that some assay technologies may have indefinite ranges of the _in-trans_ variation. These two points led us to creating the GenotypeMember class which allows us to compactly represent a very broad array of Genotype representations including but not limited to Genotypes in aneuploidy and organisms with large ploidy counts.  

Please let us know if this resolves your concerns. If not, please clarify your concerns further.

larrybabb:
> However, leaving it as non-identifiable (as it currently is) means that the `Genotype.members` array will not be sorted during serialization (but it should be).

@ahwagner I'm not happy with any of the 3 options above. Couldn't we modify the serializer/digester function to always generate a computed identifier on any "array" elements like this and use that to generate the sorted/parent digest without explicitly making the GenotypeMember truly identifiable. In others words, do the identifier work in hidden mode for the purpose of generating the next level up identifier.  I don't think we want or need to expose a computed identifier for a GenotypeMember, but it doesn't mean that one can't be created behind the scenes for the purposes of keeping things consistent as we grow.

larrybabb:
> So I think we should add some implementation guidance to accommodate this, clarifying that all Alleles / Haplotypes represented in a Genotype are presumed to be _in-trans_ on homologous sequences, and _therefore are constrained to representation on the same sequence_. We should also add to our implementation guidance that no two GenotypeMember objects can have the same Allele subject (unless we change the model to accommodate my first concern above).

I think this is the right direction. But I'm not sure I really get why having alleles at different loci as GenotypeMember's is so problematic.  If we are clear in that every GenotypeMember instance (including each "count" occurrence) is in-trans withe every other one within the genotype, then we should still be able to express genotype members as alleles from different loci. 

Although, if we want to be super precise we could make this constraint and then if someone wants to represent a compound het, then they would need to create the haplotype for the two loci containing one "same as ref" allele such that the _in-trans_ haplotype represented the compound het.  This would guarantee that someone is clearly stating that there are two changes that are _in-trans_ but not on the same copy.  I suppose this is cleaner and avoids the ability for folks to create different forms of the same equivalent genotypes.

ahwagner:
> > However, leaving it as non-identifiable (as it currently is) means that the `Genotype.members` array will not be sorted during serialization (but it should be).
> 
> @ahwagner I'm not happy with any of the 3 options above. Couldn't we modify the serializer/digester function to always generate a computed identifier on any "array" elements like this and use that to generate the sorted/parent digest without explicitly making the GenotypeMember truly identifiable. In others words, do the identifier work in hidden mode for the purpose of generating the next level up identifier. I don't think we want or need to expose a computed identifier for a GenotypeMember, but it doesn't mean that one can't be created behind the scenes for the purposes of keeping things consistent as we grow.

Taking this approach is viable, but would also require us to add explicit indices for ComposedSequenceExpressions and other `array` structures where order is meaningful. @reece and @andreasprlic do you have any opinions on a preferred strategy here?

larrybabb:
> Taking this approach is viable, but would also require us to add explicit indices for ComposedSequenceExpressions and other `array` structures where order is meaningful. @reece and @andreasprlic do you have any opinions on a preferred strategy here?

OK. so maybe if order is meaningful we don't decompose the array members into digests and sort, but when order isn't meaningful we do. Everyone wins. right?

ahwagner:
> > Taking this approach is viable, but would also require us to add explicit indices for ComposedSequenceExpressions and other `array` structures where order is meaningful. @reece and @andreasprlic do you have any opinions on a preferred strategy here?
> 
> OK. so maybe if order is meaningful we don't decompose the array members into digests and sort, but when order isn't meaningful we do. Everyone wins. right?
> 
> Honestly, it would be useful to explicitly note on any of our attributes that are arrays whether or not order is critical. Right now, it's not very transparent. I had lost track of the array structures that we decided had a meaningful order.
> 
> The other way to look at this (for consistency sake) is to say the physical message array order is never a requirement, if there is a meaningful "logical" order of things then we will clearly express that in the implementation guidance. Then we can "sort" array's with meaningful order based on the logic rules and assure that the message sender doesn't ever have to worry about getting it wrong.

I agree with this, and think we can take this a step further by adding an explicit attribute to arrays (similar to `minItems`, `items`, and other JSON Schema attributes) that tells VRS-Python and other downstream tools when to sort / not sort any given array for serialization. This is what I meant in option 2 above. To be clear, these attributes are not data properties–they would not be instantiated in messages, only present in the schema.

larrybabb:
@ahwagner  Thank you for clarifying that. I didn't see that initially. Then #2 would be my preference also.

docs/source/terms_and_model.rst
```diff
@@ -407,6 +407,12 @@ Two, three, or four total copies of BRCA1:
       "type": "CopyNumber"
     }
 
+.. _Genotype:
```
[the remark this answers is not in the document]
ahwagner: Yes! Making an issue.

schema/defs/vrs/Haplotype.rst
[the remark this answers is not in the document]
reece: Whoops. I totally misread this somehow. Disregard.

andrewyatz:
Thanks for flagging this up in the VRS call. Just to bring up some offline discussions I had with @ahwagner about this issue since we looked at the VRS algorithm for sequence collections. The underlying issue is a lack of set and array data types in JSON and JSON schema so the normalisation algorithm can't handle these two cases appropriately. We are not adopting VRS' algorithm for our arrays because of this.

I believe this means, for me, option 2 is the correct solution and using custom attributes to flag when order must be preserved.

This may lead onto an additional point whether you believe that order must also be explicit i.e. as an attribute of an array marked with sort being necessary or if it is enough to just have the elements alone. I'd personally go for an implicit order.

wesleygoar (approved):

schema/vrs-source.yaml
larrybabb: we will update the implementation notes to make sure duplicate variations in the .members attributer are not allowed.

larrybabb (approved):
looks good, assuming you get the one last item in.

docs/source/terms_and_model.rst
```diff
@@ -435,14 +432,15 @@ objects (which would otherwise be represented using symbolic shorthand).
...
+  it describes no explicit in-trans phase relationship. Therefore,
   VRS does not include an explicit entity for diplotypes. See :ref:`this note
   <genotypes-represent-haplotypes-with-arbitrary-ploidy>` for a discussion.
 * VRS makes no assumptions about ploidy of an organism or individual nor any
   polysomy affecting a locus. The `genotype.count` attribute explicitly captures the total
-  count of in-trans molecules at a genomic locus represented by the Genotype.
+  count of molecules associated with a genomic locus represented by the Genotype.
 * In diploid organisms, there are typically two instances of each autosomal chromosome,
-  and therefore two instances of sequence at a particular location. Thus, Genotypes will
+  and therefore two instances of sequence at a particular locus. Thus, Genotypes will
```
andreasprlic, line 443: I'd add that if the desire is to express a specific diplotype, it could be represented as a genotype of two haplotypes.

schema/defs/vrs/Genotype.rst
```diff
@@ -31,4 +31,4 @@ Some Genotype attributes are inherited from :ref:`Variation`.
    *  - count
       - :ref:`Number` | :ref:`IndefiniteRange` | :ref:`DefiniteRange`
       - 1..1
-      - The total number of copies of all :ref:`MolecularVariation` at this locus, MUST be greater than or equal to the sum of :ref:`GenotypeMember` copy counts. If greater than the total counts, this implies additional  :ref:`MolecularVariation` that are expected to exist but are not explicitly indicated.
+      - The total number of copies of all :ref:`MolecularVariation` at this locus, MUST be greater than or equal to the sum of :ref:`GenotypeMember` copy counts and MUST be greater than or equal to 1. If greater than the total of GenotypeMember counts, this field describes  additional :ref:`MolecularVariation` that exist but are not  explicitly described.
```
andreasprlic, line 34: I am not sure I understand why the total nr of genotypemember counts can be different from the overall count. Do you have an example where this might be needed?

ahwagner: The example we referenced when considering this use case was provided by @larrybabb, in [this example report from eMERGE](https://pubmed.ncbi.nlm.nih.gov/33930535/#&gid=article-figures&pid=fig-3-uid-2). What was found is that in these cases a heterozygous variant is reported, but no mention to the second allele (presumably reference) is given.

docs/source/terms_and_model.rst
ahwagner: Sure! Added some text for this in 40ef2af.

larrybabb (approved):
nice work. looks spot on.

PR #402, opened 2022-10-03
fix: get smoketests to pass

korikuzma:
- update python-jsonschema-objects version to >=0.4.0
- fix typo in CytobandInterval end
- Regenerate vrs yaml/json to update members to have 2 min items


ahwagner (approved):
Looks great. 👍

PR #407, opened 2022-10-24
Update validation models

### LINKED ISSUE #405: Update models.yaml
The data in `models.yaml` is invalid. 

For example, `AbsoluteCopyNumber.subject` can be a `CURIE`, `ChromosomeLocation`, or `SequenceLocation` but in `models.yaml` `AbsoluteCopyNumber.subject` is a `Gene`. 

```yaml
AbsoluteCopyNumber:
  - name: ">=3 copies APOE"
    in:
      copies:
        comparator: '>='
        type: IndefiniteRange
        value: 3
      subject:
        gene_id: ncbigene:348
        type: Gene
      type: AbsoluteCopyNumber
```

This is causing tests to fail in [this](https://github.com/ga4gh/vrs-python/pull/114) VRS-Python PR.

korikuzma:
Close #405

- Absolute / Relative Copy Number now have a CURIE as the subject
- Add test models for GenotypeMember + Genotype

### DIFF
--- a/validation/models.yaml
+++ b/validation/models.yaml
@@ -343,3 +343,135 @@ VariationSet:
       ga4gh_digest: QLQXSNSIFlqNYWmQbw-YkfmexPi4NeDE
       ga4gh_identify: ga4gh:VS.QLQXSNSIFlqNYWmQbw-YkfmexPi4NeDE
       ga4gh_serialize: '{"members":["-kUJh47Pu24Y3Wdsk1rXEDKsXWNY-68x","Z_rYRxpUvwqCLsCBO3YLl70o2uf9_Op1"],"type":"VariationSet"}'
+GenotypeMember:
+  - name: "GenotypeMember w/ Allele"
+    in:
+      count:
+        value: 1
+        type: Number
+      variation:
+        location:
+          interval:
+            end:
+              type: Number
+              value: 94842866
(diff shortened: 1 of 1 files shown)


larrybabb (approved):
i left a comment for @ahwagner and I to discuss, but nothing that should hold this up at this point in time.

ahwagner (changes requested):
I think these tests need to be changed to have a `SequenceLocation` subject–not a `Gene`. We had decided that we wanted to keep these `Location` based and not allow polymorphism. The `CURIE` is for the `Location` ID only.

larrybabb:
> I think these tests need to be changed to have a `SequenceLocation` subject–not a `Gene`. We had decided that we wanted to keep these `Location` based and not allow polymorphism. The `CURIE` is for the `Location` ID only.

I agree. I also think we should make the attribute name as precise as possible with the type. So we should seriously consider changing it to `location` instead of subject (i think).

korikuzma:
@larrybabb @ahwagner Changed the `subject` to be a `SequenceLocation` as seen [here](https://github.com/ga4gh/vrs/pull/407/commits/057342f2da8694eb57c3b2f046044c6b88a76fac).

ahwagner (approved):

PR #409, opened 2022-10-31
fix: models.yaml to take ordered property in account

korikuzma:
Updating `models.yaml` to take into account the `ordered` property in array fields. 

The docs will be updated with this new rule in #410

### DIFF
--- a/validation/models.yaml
+++ b/validation/models.yaml
@@ -427,7 +427,7 @@ GenotypeMember:
         type: Haplotype
       type: GenotypeMember
     out:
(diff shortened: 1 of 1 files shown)


reece (approved):
lgtm (but I didn't run anything... lmk if you'd like me to do that)

korikuzma:
@reece do you agree with #410? If not, we should discuss further on that issue and I should make the appropriate changes before merging this PR.

reece:
I wasn't aware of the ordered property. I commented on #410. Thanks for raising it.

ahwagner (approved):

PR #412, opened 2022-11-10
Add tests to check schema validation in models.yaml (#406)

### LINKED ISSUE #406: Improve test model validation
VRS-Python tests rely on the validation test data in VRS (`tests/validation/data/models.yaml`). However, it often appears that this does not get updated as the schema updates and causes VRS-Python tests to fail. We should improve the tests to validate the test data.

korikuzma:
Close #406

Before, we would find out that the examples in models.yaml were invalid through vrs-python tests. This added test will allow us to know in this repo

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1,10 +1,12 @@
 import json
 
+import yaml
 import python_jsonschema_objects as pjs
 from schema.helpers import pjs_filter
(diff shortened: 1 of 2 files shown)


larrybabb (approved):
looks good to me.

ahwagner (approved):

PR #414, opened 2022-12-08
change: relative copy class is now enum allowing EFO IDs

korikuzma:


korikuzma:
@ahwagner @larrybabb sorry for all the changes, `make all` was failing tests so I only committed changes that apply to the relative_copy_class update. Now ready for review.

korikuzma:
@ahwagner tests pass locally + on GH Actions now after updating the python jsonschema objects version

Issue #404, opened 2022-10-18
CopyNumber: Naming and Semantics


rrfreimuth:
As terms, "Absolute" and "Relative" are good opposites. In the proposed model, Absolute represents numerical counts, even if there is ambiguity in the count itself (in which case DefiniteRange and IndefiniteRange could be used). Relative represents levels relative to an unstated comparator (e.g., “partial loss”, “copy neutral”, “low-level gain” or “high-level gain”) but it also can represent an absolute count (“complete loss”, meaning copy count = 0).

1. Terms and values
 1a. The last example - complete loss - throws me a bit because it is an absolute count that needs no comparator but which is included in a value set expressing relative copy number. I'd like to pull that value out, but we need a way to represent it, almost like a Text Expression datatype for Absolute.
 1b. I am not sure whether "low-level gain" and "high-level gain" would be differentiable in practice, so I'd prefer to merge those concepts.

2. The difference in precision might need to be addressed. Is it possible to have a statement where there is a range of "Relative" terms (e.g., copy neutral to low-level gain)? If so, would we need a structure similar to Definite/Indefinite Range to support those terms?

3. I was thinking about whether Absolute and Relative might describe Quantitative and Qualitative concepts, respectively, but I keep going back and forth on that. Fundamentally, both are quantitative, but the values for Relative are expressed using qualitative terms (that are derived from some type of quantitative data).

In summary, my head hurts because CopyNumber is a quantitative concept that can be expressed 1) as numerical counts or as textual terms, 2) using absolute or relative frames of reference, 3) with or without ambiguity in precision.

theferrit32:
@rrfreimuth, I was just discussing something with @larrybabb that I think overlaps with your discussion here. We have a desire to specify copy number gains and losses without specifying the level of gain or loss, because that's how they are recorded in ClinVar. One option is to just select the less severe case for each, `low-level gain` and `partial loss`, but that is maybe not ideal because it is implying information that is not present in the underlying source data. But this is what we will proceed with under the current spec.

There is an EFO ontology in the EBI OLS which has these possibilities:
relative copy number variation:
- copy number gain
  - high-level gain
  - low-level gain
- copy number loss
  - complete loss
  - low-level loss
 
(https://www.ebi.ac.uk/ols/ontologies/efo/terms?iri=http%3A%2F%2Fwww.ebi.ac.uk%2Fefo%2FEFO_0030066&lang=en&viewMode=All&siblings=false)

[...]

PR #419, opened 2023-04-01
relative_copy_class  -> copy_assessment in `CopyNumberChange` example

mbaudis:
IMO `relative_copy_class`  should be `copy_assessment` in the `CopyNumberChange` example?


ahwagner:
@mbaudis great catch, agreed this needs updating–and the attribute name as well (since this is no longer described as an "assessment" per #404).

How does this change look to you?

Issue #425, opened 2023-05-16
SV-VRS wave 1 discussion: minimal representation


d-cameron:
In this first wave, we will define the minimal representation needed to encode SVs.

High-level design: two sub-components of SV-VRS. The first is a minimal representation that encodes the structural delta between the reference and a sample; the second is a higher-level grouping of these low-level building blocks. E.g. a breakpoint & CN loss can be group together and classified as a simple deletion. This groups quickly become complicated (especially in cancer) and I expect is where the most discussion will be.

This minimal representation:
- contains sufficient information to round-trip a genome in a meaningful manner (e.g. VRS-encoded CHM13 using hg38 sequences can theoretically be converted back to CHM13 fasta)
- A genome is composed of DNA segments (CN segments), and DNA adjacencies (breakpoints). This corresponds to a 'breakpoint graph' with segments as nodes, and
- Support all existing variant call types
- Supports ambiguity
  - intrinsic (i.e. homology at a breakpoint)
  - calling technology imprecision (e.g. optical mapping / CN resolution; linked reads; only read-pair support)
  - unplaced segments (i.e. CN gain - already handled in VRS 1.0)
  -  Single breakend call due to mapper ambiguity on one side
- Normalized/unambiguous representation
  - Not actually possible in practice.
  - I'll raise this as it's own issue to split out the discussion on that

# Minimal model

## Breakend
Minimal: {sequence, pos, orientation}

Example: {chr12, 1000, ConnectedAfterPosition }

### Ambiguity 
As above but instead of an intrinsic sequence-based ambiguity, it's a caller-based margin of error.
  - Which alignment convention: left/right/centre/most likely/weighted likelihood/undefined?
  - Left/right breakend margins can be different

## Breakpoint
Minimal: {breakend1, breakend2}

Technically speaking the breakends are unorder but we can define a sort order to ensure id stability.

### inserted sequence:
- can be:
  - literal sequence
  - derived sequence
     - (i.e. if the donor site can be identified)
     - Does VRS support calling variants within this derived sequence (e.g. a SNV in a translocated LINE)
- left/right flanking literal sequences (e.g. manta/DRAGEN)
   - My preference is for these 'insertions' to be represented as a pair of single breakends
- An estimated gap size
   - Needed to support BioNano optical mapping gaps

### homology

[...]

Issue #365, opened 2021-12-09
Breakends and positions


ahwagner:
One concept that will need to be considered as we approach fusion & SV representation is the notion of a _breakend_, a _position_ on a sequence representing some indefinite amount of sequence to the left or right of the position.

The position is contrasted from an interval by having a single Number/Range value instead of both a start and end Number/Range value.

cmprocknow:
I think we can represent this already using IndefiniteRange, right?:

	{
		"interval": {
			"end": {
				"type": "Number",
				"value": 44908822
			},
			"start": {
				"type": "Number",
				"value": 44908821
			},
			"type": "SequenceInterval"
	},
		"sequence_id": "ga4gh:SQ.IIB53T8CNeJJdUqzn9V_JnRtQadwWCbl",
		"type": "SequenceLocation"
	}

PR #428, opened 2023-07-02
Initial SV-VRS schema update

d-cameron:
Initial SV-VRS Schema

# High-level design summary

- Breakend represent a change from reference to non-reference sequence at a given position
- Location encode any uncertainty in the position
- Breakpoint are composed on two (unordered) breakend
- Breakends can have literal sequence after the break
- SV-aware phasing is done by making Haplotype ordered and encoding the phasing in the allele ordering
- Allele remains untouched to minimise impact on the existing schema
- Zero-width Locations are allowed (to encode exact breakpoint positions)
- End-of-chromosome breakends are encoded in Breakend.terminal instead of VCF's approach of using placeholder breakends past the end of the chromosome
- CN untouched

# VCF equivalences:

SV-VRS | VCF
------|-----
Breakpoint | Two (redundant) ALT breakpoint records
Isolated Breakend | Single Breakend ALT allele notation
Terminal Breakend | Breakpoint past end of chromosome
Haplotype ordering | PSL/PSO fields
Breakend.sequence | Encoded in ALT^ for single breakends/(redundantly in) breakpoints
Breakpoint.insertion | Not supported

^ single breakend sequence must be at least one base otherwise it will be interpreted as the missing `.` allele.

# Clarifications needed

## ~~Is Location.end inclusive or exclusive?~~

~~That is, is `Location` encoded as [start, end] or [start, end)?~~

Location is start-end inclusive

## What is "non-overlapping Allele" in Haplotype defined with respect to?

Allowing duplication means that two copies of an Alelle can exist on the same molecule. For SV-VRS to work, the definition of 'non-overlapping' has to be with respect to the molecule, not w.r.t the reference. I don't think there's any problems with this subtle redefinition.

## What is the relationship between the reference and a haplotype?

The presence of an Allele/Haplotype defines a deviation w.r.t a reference. What (if anything) can be said about the reference outside of the Allele positions? Take the following example where there are 3 SNPs in a gene:

A) A SNP chip identifies and imputes phasing for SNPs at locations A, B and C. It reports a Haplotype containing three Alleles.

B) A WGS sequencing run finds the entire gene matches the reference except for these SNPs. It reports a Haplotype containing three Alleles.

C) A WES sequencing run finds the exons of the gene match the reference except for these SNPs. It reports a Haplotype containing three Alleles.

How does VRS differentiate these three scenarios? Does the CN of the gene impact this?

## How should Breakend.sequence be encoded?

- reverse complement on DivergesBefore?
- ~~reverse complement when DerivedSequenceExpression.reverse_complement=true?~~ (Not relevant: reverse_complement is not part of Location)

## Do we need a CILEN equivalent?

The breakpoint representation chosen is lossy w.r.t common variant calling types.
VCF has CILEN which encodes the range of expected lengths for simple SVs but has no equivalent for breakpoints.
This is actually a lossy representation as many variant callers
can constrain the actual location much more than anywhere in the
[(start1, end1), (start2, end2)] range.
for example, when the interval are due to homology, then the
interval widths must be the same and, for any given position
in the first breakend interval, there is only one possible position
in the second breakend interval that is possible.
Just specifying the two intervals independently as is done in this
model does not intrinsically encode this information (although for homology we can indeed define it as such since it's an unambiguous representation).

Similarly, even imprecise calls have possibilities that are less
likely than others. For example if a deletion break1 was at start1, then
break2 might be constrained so something like [start2, start2 + (end2 - start2) / 3]
because that would imply a longer deletion length that is plausible
(hence the VCF CILEN field).

Is there value in an equivalent field? Would anyone actually make any decisions differently if there was a narrower band of possible positions for an imprecise SV call?

### DIFF
--- a/schema/vrs.yaml
+++ b/schema/vrs.yaml
@@ -73,6 +73,11 @@ definitions:
     - type
     additionalProperties: false
   Haplotype:
+    # For SV-VRS 'non-overlapping' is defined with respect to the
+    # derivate chromosome
+    # This allows SNVs within duplications to be reported as they
+    # can overlap in reference coordinate space, but not on the
+    # sample molecule.
     description: A set of non-overlapping Allele members that co-occur on the same
       molecule.
     type: object
@@ -88,12 +93,39 @@ definitions:
       members:
         type: array
         minItems: 2
-        uniqueItems: true
-        ordered: false
+        # By allowing repeats and defining an order on the
+        # members within a haplotype, we get a
+        # By allowing repeats within a Haplotype, we can now represent
+        # variants within SV. Notably, we can define SNVs that
+        # occur multiple times within a duplication.
+        uniqueItems: false
+        # The ordering of members within a Haplotype defines a traversal
+        # order for the derivate chromosome. This allows phasing of
+        # breakpoints as well as the Alleles between them.
+        # This is equivalent to the VCF phase set list (PSL) field
+        # with the VCF PSO field implicitly encoded in the member
+        # ordering within the Haplotype.
+        #
+        # Is this model appropriate? It doesn't support scenarios
+        # such as knowing two SNVs within a duplication are on the same
+        # molecule but not being able to place them to the first or
+        # second copy.
+        # (Note that the current VRS model does not support this
+        # scenario as the Alleles must be non-overlapping).
+        ordered: true
         items:
           oneOf:
           - $ref: '#/definitions/Allele'
           - $ref: '#/definitions/CURIE'
+          # Allowing Breakends and Breakpoints within Haplotypes
+          # is the lowest impact change to the existing VRS model.
+          # Conceptually, one might consider these to be Alleles,
+          # but they violate many of the assumptions that the existing
+          # Allele model has. Notably, breakpoints have two locations.
+          # Is this low-impact approach the best approach or should we
+          # be redefining Allele?
+          - $ref: '#/definitions/Breakpoint'
+          - $ref: '#/definitions/Breakend'
         description: List of Alleles, or references to Alleles, that comprise this
           Haplotype.
     required:
@@ -341,7 +373,7 @@ definitions:
         - $ref: '#/definitions/Number'
         description: The start coordinate or range of the interval. The minimum value
           of this coordinate or range is 0. MUST represent a coordinate or range less
-          than the value of `end`.
+          than or equal to the value of `end`.
       end:
         oneOf:
         - $ref: '#/definitions/DefiniteRange'
@@ -349,7 +381,7 @@ definitions:
         - $ref: '#/definitions/Number'
         description: The end coordinate or range of the interval. The minimum value
           of this coordinate or range is 0. MUST represent a coordinate or range greater
-          than the value of `start`.
+          than or equal to the value of `start`.
     required:
     - end
     - start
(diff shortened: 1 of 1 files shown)


d-cameron:
## Do not merge to main

Related issues: #365 #425

larrybabb:
@d-cameron 
re; **Is Location.end inclusive or exclusive?**
That is, is Location encoded as [start, end] or [start, end)?

VRS start/end positions represent the intervals between residues they do not represent the positions of the residues themselves. This eliminates the confusion related to inclusivity and exclusivity.  [Here's one](http://dataexchange.clinicalgenome.org/allele/discussion/) of several discussions that goes over the rational for the "interval" method.

d-cameron:
> VRS start/end positions represent the intervals between residues they do not represent the positions of the residues themselves. This eliminates the confusion related to inclusivity and exclusivity.

Using interbase coordinates doesn't eliminate the ambiguity regarding inclusive or exclusive endpoints. The link you provided does though:

> defined as the subsequence that begins at the base numbered "start" and goes to the base numbered "end", **_both inclusive_**.

d-cameron:
Issue: Haplotype is defined on a single contig and SVs can traverse multiple contigs

schema/vrs.yaml
```diff
@@ -678,3 +710,117 @@ definitions:
...
+          type: Breakend
+        # TODO: how should the Breakend.sequence be interpreted?
+        # Option 1: be1 <-> seq1 <-> seq2 <-> be2
+        # That is, concatenate the sequences of both of them.
+        # Pros: no possibility of data inconsistency
+        # Cons: need to 'allocate' the sequence to one of the breakends
+        #
+        # Option2: be1 <-> seq <-> be2
+        # Pros: more intuitive interpretation
+        # Cons: possibility of data inconsistency (when seq1 != seq2)
```
Mrinal-Thomas-Epic, line 803: How would `seq` be defined in option 2?

Mrinal-Thomas-Epic: From a discussion with Daniel, the reason for putting `sequence` at the breakend-level is that we also need to support inserted sequence on terminal breakends.

Another possibility is representing terminal breakends as a breakpoint with a single terminal breakend, and moving `sequence` up to the breakpoint.

ahwagner:
> > VRS start/end positions represent the intervals between residues they do not represent the positions of the residues themselves. This eliminates the confusion related to inclusivity and exclusivity.
> 
> Using interbase coordinates doesn't eliminate the ambiguity regarding inclusive or exclusive endpoints. The link you provided does though:
> 
> > defined as the subsequence that begins at the base numbered "start" and goes to the base numbered "end", **_both inclusive_**.

This is not correct. VRS intervals use inter-residue coordinates, or interval counting, to avoid ambiguity in inclusive/exclusive behavior. If you want to convert this to a residue-counting system, it is equivalent to `.bed` files, which are 0-based, inclusive beginning, _exclusive_ end. The link Larry provided describes "both inclusive" with respect to _the Alignment method_, but VRS uses _the Interval method_ as described in that post. Using an inter-residue system simplifies how we describe indel behavior. @bmilius-nmdp has a pretty good slide deck on the topic [here](http://17ihiw.org/wp-content/uploads/2016/10/20160925_SequenceCoordinates_ASHI2016.pdf).

schema/vrs.yaml
```diff
@@ -88,12 +93,39 @@ definitions:
       members:
         type: array
         minItems: 2
-        uniqueItems: true
-        ordered: false
+        # By allowing repeats and defining an order on the
+        # members within a haplotype, we get a
```
ahwagner, line 97: trailing comment

ahwagner (commented):
Lots of interesting work here. I think we need to align on the notion of `Sequence`, `Location`, and `Variation` as used in VRS `1.x`.

greg-sharpe:
Would it be clearer, or overly specific to use the term "linker sequence" instead of just "sequence" to refer to the literal sequence expression occurring between the breakends?

https://fusions.cancervariants.org/en/latest/information_model.html#linker-sequence

PR #438, opened 2023-08-15
2.0 alpha model changes to validation/models.yaml

toneillbroad:


korikuzma (changes requested):
Should update `copy_change` to `copyChange`

PR #442, opened 2023-08-31
fix: update digest + refgetAccession patterns

korikuzma:
I think that the regex patterns should be updated for the `digest` and `refgetAccession` fields. As I've been updating tests in downstream apps (gene-normalizer), these fields did not raise an error when I thought they should.


larrybabb (approved):
+1

ahwagner (approved):

PR #444, opened 2023-09-11
update models from 2.0-alpha changes

korikuzma:
Main changes:
- `SequenceReference` digest is `referenceSequence.refgetAccession.split("SQ.")[-1]`
- Copy Number `subject` renamed to `location`

Questions:
- I'm not sure about `SequenceReference`. I currently have it so that `ga4gh_identify` and `ga4gh_serialize` return null. It's under the IDENTIFABLE TYPES sections, so it should return a response for `ga4gh_identify`. Would it just be `ga4gh:{refgetAccession}`?

### DIFF
--- a/validation/models.yaml
+++ b/validation/models.yaml
@@ -89,6 +89,8 @@ LiteralSequenceExpression:
       sequence: ACGT
       type: LiteralSequenceExpression
     out:
+      ga4gh_identify: null
+      ga4gh_digest: ECvt22ZFJ6RcLutlH07yAkd2-hNb-dwV
       ga4gh_serialize: '{"sequence":"ACGT","type":"LiteralSequenceExpression"}'
 #RepeatedSequenceExpression:
 #  -
--- a/validation/models.yaml
+++ b/validation/models.yaml
@@ -5,7 +5,9 @@ SequenceReference:
       type: SequenceReference
       refgetAccession: SQ.F-LrLMe1SRpfUZHkQmvkVKFEGaoDeHul
     out:
+      ga4gh_identify: null
       ga4gh_digest: F-LrLMe1SRpfUZHkQmvkVKFEGaoDeHul
+      ga4gh_serialize: null
 SequenceLocation:
   - name: "SequenceLocation w/ SequenceReference"
     in:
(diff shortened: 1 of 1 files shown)


larrybabb:
@korikuzma i believe the `SequenceReference` should return exactly the same value that `Sequence` would return.

korikuzma:
> @korikuzma i believe the `SequenceReference` should return exactly the same value that `Sequence` would return.
> 
> SequenceReference is a proxy for Sequence.

@larrybabb can you suggest changes in the file?

ahwagner (approved):

ahwagner:
> Main changes:
> 
> * `SequenceReference` digest is `referenceSequence.refgetAccession.split("SQ.")[-1]`
> * Copy Number `subject` renamed to `location`
> 
> Questions:
> 
> * I'm not sure about `SequenceReference`. I currently have it so that `ga4gh_identify` and `ga4gh_serialize` return null. It's under the IDENTIFABLE TYPES sections, so it should return a response for `ga4gh_identify`. Would it just be `ga4gh:{refgetAccession}`?

You are correct to have those methods return `null`. There is no need to serialize as there is a custom digest method. There is no identifier and it should not be under identifiable types. See the schema [here](https://github.com/ga4gh/vrs/blob/375cb515876779936c302266de9a60b4b907c8d3/schema/vrs-source.yaml#L393-L395):

```
  SequenceReference:
    maturity: Alpha
    inherits: ValueObject
```

This class should inherit ValueObject, not Ga4ghIdentifiableObject.

korikuzma:
@ahwagner  do you have thoughts on the `SequenceReference` example? 

I'm thinking it should be:

```yaml
    in:
      type: SequenceReference
      refgetAccession: SQ.F-LrLMe1SRpfUZHkQmvkVKFEGaoDeHul
    out:
      ga4gh_digest: F-LrLMe1SRpfUZHkQmvkVKFEGaoDeHul
      ga4gh_identify: ga4gh:SQ.F-LrLMe1SRpfUZHkQmvkVKFEGaoDeHul
      ga4gh_serialize: null
```

korikuzma:
@ahwagner made small change to LSE and explicitly had ga4gh_serialize/ga4gh_identify return null for SequenceReference

ahwagner (approved):
All looks good to me with the caveat that I did not verify hash correctness. Let me know if this is needed.

korikuzma:
> All looks good to me with the caveat that I did not verify hash correctness. Let me know if this is needed.

I run through vrs-python and did not make changes to `sha512t24u`. I'll merge. Thanks for the quick PR review @ahwagner and @larrybabb

Issue #437, opened 2023-08-14
Copy Number Count text and json example do not match


korikuzma:
https://github.com/ga4gh/vrs/blob/c4595ae173e10195e6a7157364a6edb5a7276283/docs/source/terms_and_model.rst?plain=1#L388-L407

If we want to keep the json, we should update  the text to `3 or more copies of APOE`.

 If we want to keep the text, the json should be:

```json
    {
      "copies": {
        "type": "DefiniteRange",
        "min": 2,
        "max": 4
      },
      "subject": {
        "gene_id": "ncbigene:672",
        "type": "Gene"
      },
      "type": "CopyNumberCount"
    }
```

korikuzma:
@sivangbagri go for it

PR #454, opened 2023-11-29
Fixed copy number text and json difference issue by changing json (issue #437)

sivangbagri:


docs/source/terms_and_model.rst
```diff
@@ -387,18 +387,18 @@ as an :ref:`Allele` with a :ref:`RepeatedSequenceExpression`.
 
 **Examples**
 
-Two, three, or four total copies of BRCA1:
```
larrybabb, line 390: I believe we intended to use `BRCA1` and not `APOE` in this example, just so that it would be more easily distinguished from the prior example which is an `APOE` Haplotype. Let's leave this as `BRCA1`

docs/source/terms_and_model.rst
```diff
@@ -387,18 +387,18 @@ as an :ref:`Allele` with a :ref:`RepeatedSequenceExpression`.
...
       "copies": {
-        "comparator": ">=",
-        "type": "IndefiniteRange",
-        "value": 3
+        "type": "DefiniteRange",
+        "min":2,
+        "max":4
       },
       "subject": {
-        "gene_id": "ncbigene:348",
```
larrybabb, line 401: Let's leave the ncbigene value for `BRCA1` and keep this example about that gene.

docs/source/terms_and_model.rst
```diff
@@ -387,18 +387,18 @@ as an :ref:`Allele` with a :ref:`RepeatedSequenceExpression`.
...
+Two, three, or four total copies of APOE:
 
 .. parsed-literal::
 
     {
       "copies": {
-        "comparator": ">=",
-        "type": "IndefiniteRange",
-        "value": 3
+        "type": "DefiniteRange",
```
larrybabb, line 396: +1 good catch. `DefiniteRange` with a min/max of 2/4 is correct. thanks.

larrybabb (changes requested):
@sivangbagri thank you for the PR. I agree with your correction, but I think we should keep the example based on the gene `BRCA1`. The prior Haplotype section uses `APOE` in it's example, and I believe the original author intended to use a different gene so at to make it clear it had no connection to the prior example. Please correct and I will approve and merge.

PR #466, opened 2024-03-01
RLE Schema and Normalization Algorithm

### LINKED ISSUE #426: ReferenceLengthExpression: Documentation
@ahwagner In today's VRS call I brought up the question of how one would go from a SPDI nomenclature to a VRS SeqLocation that had a `ReferenceLengthExpression` state.  

I referenced the following clinvar variation example
```
variationID: 1802409
name: NM_018406.7(MUC4):c.7701_7702insTCAGTATCCACAGGTCATGCCACCCCTCTTCATGTCACCAACACTTCC (p.Ser2567_Ala2568insSerValSerThrGlyHisAlaThrProLeuHisValThrAsnThrSer)
type: Indel
spdi: NC_000003.12:195783878:GGAAGTGT:GGAAGTGTTGGTGACATGAAGAGGGGTGGCATGACCTGTGGATACTGAGGAAGTGT
b38_hgvs: NC_000003.12:g.195783886_195783887insTGGTGACATGAAGAGGGGTGGCATGACCTGTGGATACTGAGGAAGTGT
del_spdi: GGAAGTGT
ins_spdi: GGAAGTGTTGGTGACATGAAGAGGGGTGGCATGACCTGTGGATACTGAGGAAGTGT
```
In this example the 9th character in the `ins_spdi` seq is a T which is not the natural expansion of the `del_spdi`....

The repeating sequence in the `ins_seq` string is `GGAAGTGTTGGTGACAT` so if we put in `56` for the the RLE on a SeqLocation like the following

```
start: 195783878,
end: 195783886,
state: 56
```
How would the start/end impact the recreation of the inserted sequence?

I assumed that `56` would mean that we would take the original start-end ref and start repeating it out until we truncated it at 56. In this case that would create the sequence

`GGAAGTGTGGAAGTGTGGAAGTGTGGAAGTGTGGAAGTGTGGAAGTGTGGAAGTGT`

which is not the right answer for this normalized variation

To date, we've been able to go directly from SPDI nomenclature to VRS since we would simply take the spdi `position` as the `loc.start` and add the length of the `del_seq` to get the `loc.end` and then assume the `ins_seq` of the spdi nomenclature is the `loc.state` LiteralSequenceExpression.  

It seems like now we will need to re-normalize the SPDI in order to identify when the use case for `RefLenExpr` is appropriate.

Is this a big drawback? Or should we consider whether SPDI should include the `RefLenExpr` state in it's nomenclature so that we stay in sync with SPDI?

ahwagner:
addresses RLE schema and normalization algo for #426.

RLE information model documentation will be handled in a separate PR.


larrybabb (approved):
+1

Issue #480, opened 2024-04-12
VRS 2.0 Documentation Update


ahwagner:
- [x] Reorganize Index (#482)
- [x] Update Allele (#484)
- [x] Update SequenceLocation (#483)
- [x] Create SequenceReference (#485)
- [x] Create Adjacency (#488)
- [x] Update Haplotype to CisPhasedBlock (#490)
- [x] Review DerivativeMolecule (#491)
- [x] Create Terminus
- [x] Update LiteralSequenceExpression (585b385)
- [x] Create ReferenceLengthExpression (#426)
- [x] Create LengthExpression (f56b112)
- [x] Update serialization
- [x] Update prefix table (83743a2)
- [x] #531
- [x] Update Appendices
- [x] #532
- [x] Update introduction
- [x] Create Quick Start (point to VRS-Python notebooks)

korikuzma:
ayo 🚀

PR #482, opened 2024-04-12
Addresses epic #480 task "Reorganize Index"

ahwagner:
Linking #480


larrybabb (commented):
Looks good to me. I'm not sure about 'Profiles'. I thought that was a VA thing.

PR #484, opened 2024-04-17
Addresses epic #480 task "Update Allele"

ahwagner:


larrybabb (approved):
moving in the right direction. looking forward to the next one.

Issue #429, opened 2023-07-05
SV VRS Wave 1: Ambiguity due to intrinsic sequence-based homology


Mrinal-Thomas-Epic:
## Problem statement
As Daniel laid out in issue #425, one of our goals for SV-VRS wave 1 is to come up with a simple model for low-level structural changes from a reference sequence. Part of this goal includes support for multiple kinds of ambiguity, including intrinsic sequence-based homology (i.e., ambiguous cases where the two sequences on either end of a breakpoint are identical).

Daniel drew up the following example of sequence-based homology in his 6/1 SV SIG presentation:
![image](https://github.com/ga4gh/vrs/assets/126519671/69b9971e-2a44-4a84-bd23-ebe61c3b1ff5)

There are two possible approaches we are considering to handle this ambiguity and would like to get feedback on which one is preferred.

## Option 1: Extending VRS style ambiguity representation to breakpoints
Starting with the simple breakend model from #425 (which excludes property data types, since those will be decided later):
```json
"Breakend": {
  "properties": [
    "sequence",
    "pos",
    "orientation"
  ]
}
```
We can create a breakpoint model that includes a field `ambiguous_nucleotides` for a sequence of nucleotides that are considered ambiguous due to homology.
```json
"Breakpoint": {
  "properties": [
    "breakend_1",
    "ambiguous_nucleotides",
    "breakend_2"
  ]
}
```

Then, we can use a simple VOCA-like algorithm to take a potentially over-precise Breakpoint representation and produce the precision-corrected representation.

<details>
<summary>Rough Algorithm</summary>

```python
from bioutils.sequences import reverse_complement

class Breakend:
    def __init__(self, sequence: str, pos: int, orientation: str) -> None:
        self.sequence = sequence
        self.pos = pos
        self.orientation = orientation

class Breakpoint:
    def __init__(self, breakend_1: Breakend, breakend_2: Breakend, ambiguous_nucleotides: str) -> None:
        self.breakend_1 = breakend_1
        self.breakend_2 = breakend_2
        self.ambiguous_nucleotides = ambiguous_nucleotides


def normalize_breakpoint(breakpoint:Breakpoint):
    be_1 = standardize_breakend_orientation(breakpoint.breakend_1, "connectedAfterPos")
    be_2 = standardize_breakend_orientation(breakpoint.breakend_1, "connectedBeforePos")
    ambiguous_nucleotides = breakpoint.ambiguous_nucleotides

    # shift left
    dist_left = 0
[...]

PR #488, opened 2024-04-19
Epic #480 "Create Adjacency"

ahwagner:
addresses #480 "Create Adjacency"
related to #429


Mrinal-Thomas-Epic:
Don't we still need the homology flag to distinguish between imprecision due to assay limitations vs equivalence of sequences on either side of the adjacency?

ahwagner:
I don't know; do we? The homology flag was removed a while ago, so I don't remember why exactly that change was made.  I understand the two scenarios you are describing, but is this distinction important? I haven't explored this deeply, but it seems to me that you can infer when each of these situations are taking place from the sequence content; and if you cannot infer it, that it doesn't make a difference. Maybe a good conversation topic for Connect.

For now, I will add it back in as an optional property, and will open #489 for discussion and use cases on this topic.

PR #498, opened 2024-06-11
Change IndefiniteRange.value to be an integer

theferrit32:
See https://github.com/ga4gh/va-spec/issues/129

### DIFF
--- a/schema/defs/vrs/IndefiniteRange.rst
+++ b/schema/defs/vrs/IndefiniteRange.rst
@@ -19,7 +19,7 @@ A half-bounded range of numbers represented as a number bound and associated com
       - 1..1
       - MUST be "IndefiniteRange"
    *  - value
-      - number
+      - integer
       - 1..1
       - The bounded value; inclusive
    *  - comparator
--- a/schema/vrs-source.yaml
+++ b/schema/vrs-source.yaml
@@ -670,7 +670,7 @@ definitions:
         default: "IndefiniteRange"
         description: MUST be "IndefiniteRange"
       value:
-        type: number
+        type: integer
         description: The bounded value; inclusive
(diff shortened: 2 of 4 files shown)


theferrit32:
See https://github.com/ga4gh/va-spec/issues/129

larrybabb:
@theferrit32 can you shed more light on why this was needed?

theferrit32:
Having this field (`IndefiniteRange.value`) be an `integer` to match the type of the `Number.value` field makes the graphQL query for those objects a little nicer because you can just ask for a field called `value` of type `integer` and it works for both `IndefiniteRange` and for `Number`

See item (1) here: https://github.com/broadinstitute/gnomad-browser/issues/1472#issue-2223648155

Here in the graphQL query if I want both IndefiniteRanges and Numbers to be returned in the same response I have to alias one of those `value` fields to have a different name, since the types don't match:
https://gist.github.com/theferrit32/df380899e9974cb885c2780cfa82ce7a#file-gnomad_vrs_va-graphql-L9

larrybabb:
@ahwagner I will defer to you on this one. If we merge this I think we may need a 1.4 release (unless you think it is acceptable to backport this into 1.3?). Please provide your perspective and proceed with whichever path you see as proper.

dazzariti:
@larrybabb will review, remove Andreas as reviewer

ahwagner:
It seems to me like this would be best as a patch version. In JSON Schema, an `integer` is just a constrained JSON `number`. The use of a non-integer `number` does not conform to the information content we expect, so this shouldn't be breaking in any way, just clarifying, and therefore appropriate for patch level.

ahwagner (approved):

PR #507, opened 2024-06-30
Inherit vrs from domain entity

larrybabb:
change VRS ValueObject to inherit from DomainEntity instead of Entity


ahwagner:
🤔 I wonder if there are examples where we would use mappings in VRS? Expressions allow for creating string-based representations in different syntaxes, and it seems like the notion of a concept map directly contrasts with the notion of a value object. Maybe we moved too quickly on this.

PR #510, opened 2024-07-02
chore: remove todo in vrs-source.yaml

### LINKED ISSUE #502: remove commented out code in vrs-source.yaml
Similar to [this](https://github.com/ga4gh/cat-vrs/issues/26) issue in cat-vrs.

vrs-source.yaml contains a bit of commented out code and I think it should be deleted.

There are also places, such as [this](https://github.com/ga4gh/vrs/blob/16a8f5bb6ebe0933e52d07ab170ea072774d5a39/schema/vrs/vrs-source.yaml#L585-L605), where we might need it later. In these situations, it's not always clear why code is commented out or what changes need to be made in order for it to be uncommented out.

I am proposing to remove out commented code to help keep our codebase clean. If there are places where code is commented out that we might need later, I would suggest creating an issue for it to address later.

korikuzma:
close #502

@larrybabb removed the majority! Thanks

### DIFF
--- a/schema/vrs/vrs-source.yaml
+++ b/schema/vrs/vrs-source.yaml
@@ -327,8 +327,6 @@ $defs:
           A boolean indicating whether the molecule represented by the sequence is circular (true) or linear (false).
     required: ["refgetAccession"]
 
-      #TODO: implementation guidance: digest property is RefGet TRUNC512
-
   # =============================================================================
   # NON-IDENTIFIABLE TYPES
   # These types have a `type` attribute


schema/vrs/vrs-source.yaml
```diff
@@ -327,8 +327,6 @@ $defs:
           A boolean indicating whether the molecule represented by the sequence is circular (true) or linear (false).
     required: ["refgetAccession"]
 
-      #TODO: implementation guidance: digest property is RefGet TRUNC512
```
korikuzma, line 330: @larrybabb Let me know if I need to make a new issue for this or if we're good

larrybabb: your good

larrybabb (approved):

Issue #479, opened 2024-03-29
Should SequenceReference be a Ga4ghIdentifiableObject?


Mrinal-Thomas-Epic:
I saw that `SequenceReference` has an alternate digest convention, but I think it doesn't actually define a `digest` property or inherit from a type that defines one.

ahwagner:
It should have a `keys` parameter for use with Ga4ghIdentifiableObject, but it does not create its own computed identifier. Convention should be updated to reflect #465.

larrybabb:
@ahwagner I need some clarification on this.  The `keys` parameter is used to define the attributes of a value object that are then ingested into the algorithm to generate a new computed key (with json structure intact of course).  So I don't think we want `SequenceReference` to have a new computed digest, but instead have some specialized logic that passes through the digest of the `Sequence`.  If we are saying that we DO want `SequenceReference` to have a different digest then I would definitely have the `type` attribute be included so it works like all other ValueObjects.  If not, and we don't want to have `id` collision by using the same digest for both `Sequence` and `SequenceReference` (maybe better named SequencyProxy?) then we should maybe just re-prefix it so that we have `ga4gh:SQ.<seq digest>` for `Sequence` and `ga4gh:SR.<seq digest>` for `SequenceReference`.

korikuzma:
See [this](https://github.com/ga4gh/vrs/commit/fd4a7b72c9e957fe1fa72f48e27a24119ab9e5d6) commit where `keys` was added to `SequenceReference`

larrybabb:
@ahwagner ... @korikuzma and I decided to revert the April commit where you removed the `assigned` value and replaced it with `keys`.  This change would cause `SequenceReference` to get a new computed id and that would propogate up to `SequenceLocation` and above. We do NOT want any changes to identifiers as a result of whatever option we land on here. So for now, we are reverting until we can sort this out.

korikuzma:
@larrybabb See #514 . I think this helped resolve my confusion.

korikuzma:
Okay, I think we're good but it might be nice to meet about this once @ahwagner gets back to make sure we're all on the same page.

ahwagner:
This was discussed at length in #465 and there was clear consensus we did not want special cases. Reverting that work goes against the consensus decision.

I am on mobile so might be misunderstanding. Please read through above discussion and confirm that you still want to revert to prior model; if so I will revisit when I return.

korikuzma:
[...]

PR #514, opened 2024-07-03
fix: resolve `SequenceReference` digest keys

korikuzma:
Resolve issue in #479

* Updates description for Value Object wrt serialization / digests
* Add missing `type` to `ga4ghDigest.keys`

### DIFF
--- a/schema/vrs/json/SequenceReference
+++ b/schema/vrs/json/SequenceReference
@@ -5,7 +5,10 @@
    "type": "object",
    "maturity": "draft",
    "ga4ghDigest": {
-      "assigned": true
+      "keys": [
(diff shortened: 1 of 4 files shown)


schema/vrs/json/SequenceReference
```diff
@@ -5,7 +5,10 @@
    "type": "object",
    "maturity": "draft",
    "ga4ghDigest": {
```
korikuzma, line 7: I don't love the name for this, but that can be another discussion for a later time. I think digest in the name can be confusing since value objects do not compute ga4gh digests.

larrybabb (approved):

PR #520, opened 2024-07-31
fix!: efo -> EFO

korikuzma:
* @cmungall pointed out the case difference in github.com/ga4gh/cat-vrs/issues/39
* copyChange now follows the convention preferred by EMBL-EBI as the EFO authority
* CopyNumberChange variation computed identifiers will change


schema/vrs/def/CopyNumberChange.rst
```diff
@@ -3,15 +3,15 @@
 An assessment of the copy number of a :ref:`Location` or a :ref:`Gene` within a system (e.g. genome, cell, etc.) relative to a baseline ploidy.
 
     **Information Model**
-    
+
```
jsstevenson, line 6: rip

jsstevenson (approved):

larrybabb (approved):

PR #526, opened 2024-08-09
Correct root inheritance

larrybabb:
* Delete root ValueObject which inherited from DomainEntity
* Replace ValueObject references with Entity inheritance to restore proper hierarchy


larrybabb:
@ahwagner please give this a quick review and let's merge it.
@korikuzma alex and i discussed this change and I will create a PR on vrs-python that i'd like you to review and validate if you can (tbd)

PR #556, opened 2024-10-09
Add constraint to items in adjoinedSequences

### LINKED ISSUE #474: Create schema-level constraint for start/stop on SequenceLocation within Adjacencies
Currently, the `SequenceLocation` within an `Adjacency` is expected to have either `start` or `stop` populated, but not both. The constraint that both should not be populated is not specified anywhere in the schema, which could lead to confusion (e.g., discussion #471).

Mrinal-Thomas-Epic:
Fixes #474

### DIFF
--- a/schema/vrs/vrs-source.yaml
+++ b/schema/vrs/vrs-source.yaml
@@ -521,6 +521,8 @@ $defs:
           oneOf:
           - $ref: "/ga4gh/schema/gks-common/1.x/data-types/json/IRI"
           - $ref: "#/$defs/Location"
+          not:
+            required: ["start", "end"]
         description: The terminal sequence or pair of adjoined sequences that defines in the adjacency.
         minItems: 2
         maxItems: 2
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -1,7 +1,9 @@
 from config import test_path, examples_path
 import yaml
 from config import validator
-
+from pytest import raises
+from jsonschema.exceptions import ValidationError
+from contextlib import nullcontext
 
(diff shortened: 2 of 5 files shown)


tests/test_examples.py
```diff
@@ -10,7 +12,9 @@ def test_examples():
         with open(examples_path / test['test_file']) as datafile:
             data = yaml.safe_load(datafile)
         class_validator = validator[test['definition']]
+        
         try:
-            assert class_validator.validate(data) is None
+            with raises(ValidationError) if test.get("shouldValidationFail") else nullcontext():
+                assert class_validator.validate(data) is None
```
ahwagner, line 18:
```suggestion
            if test.get("shouldValidationFail"):
                with raises(ValidationError):
                    assert class_validator.validate(data) is None
            else:
                assert class_validator.validate(data) is None
```

TIL about `contextlib.nullcontext`! I recommend rewriting this a little more verbosely for later understandability, above is an example of what this might look like. If you were to do so, also consider dropping the contextlib import statement. This is a style preference only.

schema/vrs/vrs-source.yaml
```diff
@@ -521,6 +521,8 @@ $defs:
           oneOf:
           - $ref: "/ga4gh/schema/gks-common/1.x/data-types/json/IRI"
           - $ref: "#/$defs/Location"
+          not:
+            required: ["start", "end"]
```
ahwagner, line 525: Perfect.

ahwagner (approved):
LGTM. One minor stylistic change suggested, but not required.

PR #573, opened 2024-10-28
Seqref updates

### LINKED ISSUE #567: SequenceReference molecule type
We should add molecule type for SequenceReference: Genomic, Transcript, Protein

### LINKED ISSUE #568: SequenceReference literal
We have a use case for the MAVE mapping data where we want to describe a sequence reference _with_ a corresponding literal sequence.

We should add a `sequence` field (analogous to the same concept in `sequenceLocation`) that allows us to explicitly describe the literal sequence.

ahwagner:
Addresses #567 and #568. For discussion.


schema/vrs/vrs-source.yaml
```diff
@@ -343,6 +343,17 @@ $defs:
         enum:
           - aa
           - na
+      sequence:
+        type:
+          $ref: "#/$defs/SequenceString"
+        description: >-
+          A :ref:`SequenceString` that is a literal representation of the referenced sequence.
+      moleculeType:
```
ahwagner, line 358: @cmprocknow concern about "molecule type"; perhaps "molecularContext" would be cleaner.

schema/vrs/vrs-source.yaml
```diff
@@ -343,6 +343,17 @@ $defs:
...
+        type: string
+        enum:
+          - genomic
```
ahwagner, line 363: general discussion on genomic expressed concerns, possible alternate is "template"

ahwagner: except "template" is a total non-starter. There are genomic variants that are not necessarily going to be involved as template sequence (e.g. variants in regulatory elements)

ahwagner: Group consensus: we should cover the same concept types that RefSeq uses.

schema/vrs/def/SequenceReference.rst
```diff
@@ -82,6 +82,16 @@ Some SequenceReference attributes are inherited from :ref:`gks-core:Entity`.
       - string
       - 0..1
       - The interpretation of the character codes referred to by the refget accession, where "aa" specifies an amino acid character set, and "na" specifies a nucleic acid character set.
+   *  - sequence
+      - 
+      - {'$ref': '#/$defs/sequenceString'}
+      - 0..1
+      - A :ref:`SequenceString` that is a literal representation of the referenced sequence.
```
larrybabb: This should be lowercase `sequenceString`...

schema/vrs/def/SequenceReference.rst
```diff
@@ -82,6 +82,16 @@ Some SequenceReference attributes are inherited from :ref:`gks-core:Entity`.
...
+   *  - moleculeType
+      - 
+      - string
```
larrybabb, line 92: MappableConcept? or are you good with an enum here? I defer to your judgement on this. I almost think we should avoid introducing MappableConcept into VRS unless it really needs to happen.

schema/vrs/vrs-source.yaml
larrybabb: I think we can sort this out during public review.

schema/vrs/vrs-source.yaml
larrybabb: lgtm

schema/vrs/def/SequenceReference.rst
larrybabb: forget this. enum works for this one.

larrybabb (changes requested):

schema/vrs/def/SequenceReference.rst
ahwagner: good catch!

schema/vrs/vrs-source.yaml
ahwagner: RefSeq calls this Molecule Type, so seems appropriate to keep.

Issue #569, opened 2024-10-26
Add digest and key info to RTD docs


ahwagner:
Currently the `ga4gh` field info may be found in schema, but not in docs. Update the metaschema processor tooling to include this information when available.

korikuzma:
PR in MSP is here: https://github.com/ga4gh/gks-metaschema/pull/29

larrybabb:
@korikuzma fixed this with b16 of the MSP. I was able to apply it on Nov.07.2024 to the GKS stack and VRS 2.x specifically.

PR #584, opened 2024-11-07
udpate SM & MSP

larrybabb:
* MSP updates now show `keys` section and `flags` column on <class>.rst generated files. 
* SM for core associates latest with same MSP update + new changes to EvidenceLine.


larrybabb:
fixed #569

PR #621, opened 2024-12-20
cicd: add precommit checks to GitHub actions

jsstevenson:
(low priority)

I noticed in #622 that the latest commit on this branch seemed to be in violation of some precommit hooks, so small changes to the schema source were introducing downstream changes to a bunch of unrelated def/json files. 

This PR 

1. catches up on those changes and
2. adds a precommit check to github actions to help contributors remember to run appropriate make commands/formatting hooks in their editors (or to install precommit themselves).

Higher order question as pointed out by @korikuzma in #622 : are all of these autogenerated changes correct (though it's also plausible that I'm messing it up somehow)?


korikuzma (commented):
So I'm not sure if we want to merge this if it results in removing GA4GH Digest documentation (for example, in `Adjacency`). Maybe for now we remove `update-json-def-files` from the action until we figure out what's happening?

korikuzma:
> So I'm not sure if we want to merge this if it results in removing GA4GH Digest documentation (for example, in `Adjacency`). Maybe for now we remove `update-json-def-files` from the action until we figure out what's happening?

I tried running `make all` locally and none of the GA4GH Digest docs were removed

jsstevenson:
I see the problem. I just reverted + re-ran and I think it now captures what you're referencing.

korikuzma (approved):

ahwagner (approved):

PR #622, opened 2024-12-20
chore: fix line endings, white space, typos

jsstevenson:
also includes a bunch of additional changes being made by precommit (see #621)

### DIFF
--- a/schema/vrs/def/Adjacency.rst
+++ b/schema/vrs/def/Adjacency.rst
@@ -6,21 +6,6 @@
 
 The `Adjacency` class represents the adjoining of the end of a sequence with the beginning of an adjacent sequence, potentially with an intervening linker sequence.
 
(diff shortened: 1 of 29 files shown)


korikuzma:
Hm. This PR points out that `type` is being removed from ga4gh digest generation... I don't think this is what we want since this would change all computed identifiers. @ahwagner @larrybabb 

Update: I wonder if it's something with MSP since `type` is a key for ga4gh identifiable objects: https://github.com/ga4gh/vrs/blob/2.0.0-ballot.2024-11/schema/vrs/vrs-source.yaml#L37-L39

jsstevenson:
@korikuzma I'm closing this for now -- #621 covers everything I noticed here except for a typo ("grater" -> "greater"). Can reopen to fix the typo later.

Issue #603, opened 2024-11-26
Automate examples in docs


ahwagner:
@jsstevenson absolutely. this is a "quick and dirty" plan to get it done for the public comment this week. I will add a task to automate and keep things in sync for the long term.

_Originally posted by @larrybabb in https://github.com/ga4gh/vrs/issues/601#issuecomment-2498025160_

PR #624, opened 2025-01-09
Update CONTRIBUTING.md to reference the correct branch (2.x)

srinitha709:


korikuzma (changes requested):
Thanks for the PR. This issue only requires updating the `CONTRIBUTING.md` file. It also looks like you were also updating #603 which should be done in a separate PR.

korikuzma (changes requested):
.gitignore + docs/source/conf.py might've been accidentally deleted

PR #633, opened 2025-02-19
feat: add beacon syntax

jsstevenson:
(low priority)

Following up from https://github.com/ga4gh/vrs-python/issues/478#issuecomment-2666896871. It seems like either VRS should support this Beacon syntax style, or VRS-Python shouldn't translate from Beacon.

also -- use of "should" in the description seems incorrect for an enum. I can revert that change if it's intentional though.


korikuzma (commented):
This is fine to me, but I defer to the co-leads

larrybabb (approved):

PR #663, opened 2025-12-03
Intronic Positions

ahwagner:


larrybabb (approved):
I looked through the changes and only considered changing the enumerations for the offset direction from `left` and `right` to some other terms more aligned with sequence direction. Maybe `5prime` and `3prime` to indicate the sequence endpoint direction or `positive` and `negative` to indicate the relation to the sequence numbering? 

In any case we can change that later if it is worthwhile. I think for now we should get this ina release so folks (like me) can try it out.

Nice job @ahwagner !!!
👍

PR #664, opened 2025-12-08
chore: bump gks-core from 1.0.0 to 1.1.0

korikuzma:
Once this is merged, I will:

1. Cherry-pick this commit into branch `2.0` (not sure if I have permissions to force push, so will prob have to create a new PR)
2. Create a VRS 2.0.2 release (which will create the VRS 2.0.2 tag)

PR #665, opened 2025-12-09
chore: bump gks-core from 1.0.0 to 1.1.0 (#664)

korikuzma:
Cherry-picked 62870585b98d217dc93c7f8bc6c1a5dfd0f0cd89 from #664 

Once this is merged, I will create a 2.0.2 release (which will create the 2.0.2 tag) off of the 2.0 branch.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "submodules/gks-core"]
 	path = submodules/gks-core
 	url = https://github.com/ga4gh/gks-core.git
-	branch = 1.0
+	branch = 1.1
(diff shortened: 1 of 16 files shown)


larrybabb (approved):

larrybabb:
@korikuzma @ahwagner if we make a minor change to a submodule doesn't that automatically mean all dependent modules need to be revised with a minor version update?

If adding a new class like `ConceptSet` to gks-core causes it to go from 1.0 to 1.1 then wouldn't the introduction of this new class also be a minor change to vrs, cat-vrs, etc..? Making VRS only a patch release implies that something broken was fixed, not that something new was made available (I think). 

We should consider how this is handled and document it in the release management policy doc for future consistency.

PR #671, opened 2025-12-27
docs: fix refs, add warnings check to CI/CD

jsstevenson:
* Add a GitHub action that builds docs with `SPHINXOPTS="-w"`, which means any warnings show up as a failed test. Maybe this will end up being annoying, but I think it'd be good to see. 
* Fix some broken refs. When a `:ref:` is broken, it just renders as the literal (i.e. ``:ref:`gks-development``` -> `gks-development`) which looks kinda bad. In several cases, there were refs to pages that no longer existed in 2.x.
* Remove "what's next" section in the "appendices -> example" page. This included several broken refs and referred to an appendix page that didn't exist anymore. Insofar as it provided relevant information, I didn't think this was a particularly good place to be housing it, so I just wiped the whole section.
* Change several textual and hyperlink references to the VRS 1.x `Sequence` that were still floating around. This one is kind of tricky. VRS 2.x more or less swaps out `Sequence` for `sequenceString`, as far as I can tell, but in some cases the language seemed more like it was describing a physical biological sequence and not a string-encoded version of it, so I tried my best. But I could easily just make the more naive switch of `sequence` to `sequenceString` everywhere as well.
* Add the "inter-residue coordinates" section from VRS 1.3 design decisions back. I am not sure if this was removed on purpose or not, but I think it's helpful and there is also a pretty specific reference to it in the "examples" page that deserves additional explanation.
* Fix a reference that was pointing to `Code.rst` but should've pointed to `code.rst`. This actually worked fine both on my local machine and in the readthedocs build but failed on the github action runner. (I thought the latter 2 both ran on ubuntu so I dunno what happened there)
* Remove a todo, "Describe and add a ref to an intronic variant profile". I dunno, this was the only thing I couldn't fix myself and there's a plugin (enabled 6 years ago) that turns all TODO's into warnings, which result in a failed Actions test. I think it should be converted into an issue so that it doesn't get lost.

### DIFF
--- a/.github/workflows/cqa.yaml
+++ b/.github/workflows/cqa.yaml
@@ -22,3 +22,20 @@ jobs:
       - uses: pre-commit/action@v3.0.1
         with:
           extra_args: ${{ matrix.cmd }} --all-files
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          submodules: recursive
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: 3.13
+
+      - name: install dependencies
+        run: python3 -m pip install -r .requirements.txt
+
+      - name: attempt docs build
+        working-directory: docs
+        run: make all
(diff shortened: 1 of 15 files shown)


larrybabb (approved):
@jsstevenson thank you for the great clean up work.

PR #672, opened 2025-12-27
docs: fix orphaned pages showing up in search results

jsstevenson:
If you search for a page that exists in `def/` like Allele, the search results will show you the page the definition is embedded in, and also an orphaned (ie outside of the `toctree` version of the file), showing up as "<No title>". This PR fixes the "no title" instance.

Screenshot 2025-12-26 at 9 33 44 PM


larrybabb (approved):
+1

PR #673, opened 2025-12-30
docs: auto-update copyright year

jsstevenson:
Sphinx 8.1 introduces support for `%Y` as a way to auto-update the copyright year range, if that's something we care about

### DIFF
--- a/.requirements.txt
+++ b/.requirements.txt
@@ -4,8 +4,8 @@ referencing
 ipython
 pyyaml
 ga4gh.gks.metaschema==0.3.2
-sphinx ~= 7.2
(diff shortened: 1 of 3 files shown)


ahwagner:
Hrm. This doesn't seem to render correctly on the [RTD preview build](https://vr-spec--673.org.readthedocs.build/en/673/). Some sort of config error we need to address?

jsstevenson:
Thanks @ahwagner  -- looks like there's a separate reqs file for doc generation. It's building properly now https://vr-spec--673.org.readthedocs.build/en/673/

ahwagner (approved):

PR #674, opened 2026-01-05
revert: Revert "Intronic Positions (#663)"

korikuzma:
This reverts commit 5528f89c95f39d6896fd362ad5419f7935b30be1.

@jsstevenson noticed that this was pushed to the wrong branch. @jsstevenson Thanks for pointing this out!

Since #663 modified a trial use class, this is technically a minor release and should not have gone in `2.0` (should go in `v2` and then `2.1`).


ahwagner (approved):

ahwagner:
@korikuzma LGTM. Squash and merge when you are ready.

PR #676, opened 2026-01-07
test: add validation tests for intronic positions

korikuzma:
* SequenceOffsetLocation, RelativeSequenceLocation, RelativeAllele

Tried to follow [Lucid chart](https://lucid.app/lucidchart/f641676b-2cf9-40da-93fe-c20457c1423a/edit?page=0_0#)


validation/models.yaml
```diff
@@ -96,6 +96,64 @@ SequenceLocation:
...
 
+SequenceOffsetLocation:
+  - name: NM_001385641.1:c.609+5G>A
+    in:
+      type: SequenceOffsetLocation
+      sequenceReference:
+        id: NM_001385641.1
+        type: SequenceReference
+        refgetAccession: SQ.cXRaN1IhgR6Mg9U-qafH2Be6IuQgB6Ox
+      anchor: 610
```
Mrinal-Thomas-Epic: Shouldn't this be 609, rather than 610?

Basing this on thinking through this simplified example
image

validation/models.yaml
```diff
@@ -96,6 +96,64 @@ SequenceLocation:
...
+      ga4gh_identify: null
+      ga4gh_serialize: '{"anchor":610,"anchorOrientation":"left","offsetEnd":5,"offsetStart":4,"sequenceReference":{"refgetAccession":"SQ.cXRaN1IhgR6Mg9U-qafH2Be6IuQgB6Ox","type":"SequenceReference"},"type":"SequenceOffsetLocation"}'
+  - name: NM_001385641.1:c.1058-6C>G
+    in:
+      type: SequenceOffsetLocation
+      sequenceReference:
+        id: NM_001385641.1
+        type: SequenceReference
+        refgetAccession: SQ.cXRaN1IhgR6Mg9U-qafH2Be6IuQgB6Ox
+      anchor: 1059
```
Mrinal-Thomas-Epic: Should this be 1057?

validation/models.yaml
```diff
@@ -186,6 +245,40 @@ Allele:
...
+          anchor: 1118
+          anchorOrientation: left
+          offsetStart: 12
+          offsetEnd: 13
+      relativeState:
+        type: LiteralSequenceExpression
+        sequence: A
+      baseState:
+        type: LiteralSequenceExpression
+        sequence: A
```
korikuzma: Only thing we did not review together @ahwagner @larrybabb

larrybabb: @ahwagner... `baseState` v `relativeState`? is this only different when the transcript is in the opposite direction?  So if the `LiteralSequenceExpression` is `A` in the `baseState` it would be `T` in the `relativeState`?

larrybabb (commented):
I left a comment for @ahwagner because I'm not 100% confident of the purpose of both base/relative State fields. I believe the relativeState would always be the complement of the baseState when the mapped sequence is in the opposite direction of the base sequence. 

!!! think we should consider changing the name of `relativeState` to `mappedState`. It seems like `baseState` clearly aligns with the `relativeLocation.baseSequenceLocation` so I would think that `mappedState` would align similarly with the `relativeLocation.mappedSequenceLocation`

validation/models.yaml
ahwagner: Yeah, I see this in the PR as "mappedState". I'm on the fence about it, okay either way, I think I do prefer "mappedState". We should definitely keep the class as "relativeAllele" and definitely keep the "relativeLocation" slot on the relative allele.

validation/models.yaml
```diff
@@ -96,6 +96,65 @@ SequenceLocation:
...
+      sequenceReference:
+        id: NM_005188.4
+        type: SequenceReference
+        refgetAccession: SQ.sGOtbqhneKsAZHmQ47sgcLmcYPcVTRbd
+      anchor: 1174
+      anchorOrientation: right
+      offsetStart: -1
+      offsetEnd: 0
+    out:
+      ga4gh_digest: null
```
ahwagner, line 127: We need to review digest behavior for RelativeLocation and RelativeAllele objects

korikuzma:
Possibly todo in separate branch: Update validation tests for #700 , if applicable

ahwagner (approved):

PR #675, opened 2026-01-05
feat: add Intronic Positions

korikuzma:
Original PR: #663

@jsstevenson noticed that this was pushed to the wrong branch. @jsstevenson Thanks for pointing this out! 

Since #663 modified a trial use class, this is technically a minor release and should not have gone in `2.0` (we are reverting this change in #674).

PR #680, opened 2026-02-03
docs: update gks class diagram + use draw.io

korikuzma:
* Added classes from #675 
* Moved Traversal Block out of Molecular Variation to follow schema
* Did some cleanup for consistent font color/size


ahwagner (changes requested):
@larrybabb is going to update this diagram to add relationship between SequenceExpression and Entity

larrybabb:
done.

larrybabb (approved):
+1

PR #686, opened 2026-02-13
test: add tests for ensuring correct properties

### LINKED ISSUE #683: Consider adding additional tests
When working on intron positions, we missed some things that could have been caught by additional tests. 

It would be nice to have tests for:
1. Ensure that ga4gh.inherent properties actually exist in the model (sometimes properties get renamed, but forget to update this field)
2. Ensure that the type property matches the class name (this seems to be the convention)

korikuzma:
close #683

* Ensure that ga4gh.inherent properties actually exist in the model (sometimes properties get renamed, but forget to update this field)
* Ensure that the type property matches the class name (this seems to be the convention)

### DIFF
--- a/tests/test_basic.py
+++ b/tests/test_basic.py
@@ -1,6 +1,23 @@
+import json
 from ga4gh.gks.metaschema.tools.source_proc import YamlSchemaProcessor
+import pytest
 
-from config import vrs_source_path, validator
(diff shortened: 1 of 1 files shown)


korikuzma:
Tests are expected to fail since we missed these in #663

larrybabb (approved):

PR #687, opened 2026-02-25
Thinktank examples wip

melissacline:
Our goal in today's GKS Think Tank was to provide some VRS sample data objects for each type of variant.  Thanks to the data hiding in the examples directory, most of that work was already done!  

- We added a JSON object for each YAML object, so that implementers will have something to interact with while they develop their software
- We added a README file describing what these examples are, with a pointer to the test definitions file, where many of them are described
- We added an example simple allele, a SNV, which was one variant type not yet represented in the set.  (should this also go into the test data, for consistency?)
- To give these data visibility, we added notes to the bottom of the main README, and to the Quickstart page in readthedocs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,3 +91,7 @@ tools with VRS.
 To run the smoke tests:
 
     make test
+
+## Sample Data
+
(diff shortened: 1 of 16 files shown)


melissacline:
This has been addressed as part of the Quick Start guide work in the Connect 2026 hackathon

Issue #10, opened 2017-06-12
overlaps in Interval


pnrobinson:
The following code in models.py would return false for two zero length intervals, e.g., Interval a(42,42) and Interval b(42,42).

Is this desired behavior?

def overlaps(self, other):
    assert isinstance(other, Interval)
    return self.end > other.start and other.end > self.start

reece:
Peter- sorry for not getting to this earlier.

I think this gotcha is a consequence of logic that's required for more typical cases.

First, recall these are interbase intervals. So, a(42,42) and b(42,42) are zero-width points (not width 1 sequences).

Let's consider non-zero intervals firs. Say c(5,7) and d(6,8), both length 2. The condition for overlapping is

`c.end > d.start and d.end > c.start`, which is true.

(For inclusive coordinates, this would be written as ≥)

Imagine e(7,9). Intuitively, we would not think that c and e overlap (again, interbase coordinates, so they abut at 7).

Finally, imagine moving c.start rightward to 7 (so that it becomes c2(7,7) and e.end leftward to e2(7,7). Obviously, if c(5,7) and e(7,9) don't overlap, their shrunken derivatives c2 and e2 shouldn't be seen to overlap either.

So, in effect, the condition is essentially asking whether the overlap length is non-zero. That reinterpretation is critical for thinking about whether any segments overlap, including the odd-seeming case of two zero-width points.

Does that pass the Peter logic filter?

reece:
Closing old question.

PR #691, opened 2026-04-13
2.1.0 connect 2026 #10 MNVs

jdeligt:
Adding first MNV set
@larrybabb


ahwagner:
This is stale and @larrybabb didn't want to keep it.

PR #696, opened 2026-06-24
Minor cosmetic and doc updates

larrybabb:
Random finds as I'm using the documentation. Collecting non-breaking doc or example changes here for a single PR update.


larrybabb:
based on wrong branch see #697

PR #697, opened 2026-06-24
minor cosmetic and doc updates

larrybabb:
collection minor updates to docs as i find them during use of the documentation. allows a single place to capture a number of random non-breaking updates.


larrybabb:
ugh. wrong approach. 3rd time will work.

PR #700, opened 2026-07-01
Revise normalization algorithm to select smallest factor

### LINKED ISSUE #699: docs: write RLE normalization documentation
Update documentation to support _lowest value_ repeat subunit length.

ahwagner:
Closes #699.


korikuzma (commented):
We should check to see if tests digests should be updated (probably)

ahwagner:
re: @korikuzma comment above, we should align this with #676

korikuzma:
> re: @korikuzma comment above, we should align this with #676

hey, i forgot about that pr. wow. thanks

larrybabb (approved):

PR #702, opened 2026-07-17
Fix inconsistent seed_length formatting in normalization docs

### LINKED ISSUE #649: Use consistent term for `seed_length` in allele normalization instructions
It's defined as `seed_length` but later referred to as `seed length`, making `ctrl/cmd-f` not work

https://vrs.ga4gh.org/en/stable/conventions/normalization.html#allele-normalization

Image

iharshitmishra:
Fixes #649

## Summary
The term `seed_length` was inconsistently written as "seed length" (with a space) in one instance in the normalization conventions doc, while everywhere else it uses the underscore form. This inconsistency broke Ctrl/Cmd+F search for the term on the page.

## Change
Updated the inconsistent instance to use `seed_length` consistently, matching its definition earlier in the document.

### DIFF
--- a/docs/source/conventions/normalization.rst
+++ b/docs/source/conventions/normalization.rst
@@ -57,7 +57,7 @@ whenever possible.
 Beginning with VRS 2.0, the normalization algorithm was extended to
 leverage reference-encoded variant states, providing a mechanism for
 compact representation of alleles that can be derived directly from the
-reference sequence.
+reference_sequence.
 
 LiteralSequenceExpression Alleles
 #################################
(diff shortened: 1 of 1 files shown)


korikuzma (commented):
I think if we're doing this, we may as well be consistent in the rest of the doc. For example, `reference sequence` and `alternate sequence` do not follow snake case convention.

iharshitmishra:
Updated changed reference_sequence and alternate_sequence to snake_case as suggested. Ready for another look whenever convenient.

larrybabb (commented):
@harshitmishra00182-ops Good call out. thank you

PR #711, opened 2026-08-31
docs: fix RFC 8785 section citation for control character codes

ahwagner:
`computed_identifiers.rst` cited RFC 8785 §3.2.2.1 for predefined JSON control character escape codes, but the adjacent link already pointed to §3.2.2.2, which is the correct section (§3.2.2.2 is "Serialization of Strings"; §3.2.2.1 is "Serialization of Literals" — `null`/`true`/`false`). Fixes the label to match the link.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


larrybabb (approved):
+1

PR #714, opened 2026-09-01
docs: align maturity model page with GA4GH TASC policy

ahwagner:
## What

Aligns `docs/source/appendices/maturity_model.rst` with the GA4GH TASC [Technical Specification Development](https://www.ga4gh.org/wp-content/uploads/2026/01/TASC-Technical-Specification-Development.pdf) policy, and cites that policy throughout.

## Why

The page read as though the maturity model were a GKM invention — it never referenced the GA4GH policy that actually defines it — and had drifted from that policy in several places, including one substantive rule difference on minor version increments.

## Content restored from the TASC policy

- **Alignment with GA4GH Product Development Processes** (new section), including the rule of thumb that a feature warranting a major/minor increment absent a maturity system should be annotated with a maturity level, and that validation tests / documentation appendices are not.
- **Data Class Inheritance and Property Maturity** (new section) — child data classes may not exceed the maturity of the classes they inherit from, and a class's properties may not exceed the maturity of the class itself. Neither constraint appeared on the page.
- **Communicating Maturity Level** (new section) — the JSON Schema `maturity` keyword, plus pointers to the per-class notes and the class diagram.
- The **definition of a release** as the full repository snapshot (models, upstream dependencies, docs, guidance, tests, examples).

## Divergences corrected

- Minor version increment: "Addition of optional fields to data models" → **"...at the trial use or normative level"**. This is a real rule difference, not wording.
- "releases of the entire specification" → "releases of technical specifications"; "additionally follow" → "should additionally follow".
- The Draft **Process** step said only "Follow the GKM development process", with no definition or link. It now summarizes the GA4GH four-phase product feature development process and links the VRS Discussions and Issues boards.
- **Product Leads → Product Owners**, per the TASC decision-maker definitions. The `product-leads` anchor is renamed to `product-owners`; nothing outside this page referenced it.
- Page retitled **GKM Maturity Model → GA4GH Maturity Model**.

## Also on this page

- **GKS vs GKM.** GKS is the Work Stream; GKM is the family of models (VRS, Cat-VRS, VA-Spec) within it. The page conflated the two; corrected, and the distinction is now stated in the lead note.
- The pre-release label list used un-nested `-` bullets, so each label rendered as a run-on paragraph (`connect.<YYYY>-<MM>[.<N>] - for pre-releases... - date corresponds...`). Fixed to real nested lists.
- Added the missing `(source)` link on the adoption lifecycle figure.

## Notes for reviewers

- **Base is `2.1`, not `main`.** `ga4gh/vrs` has no `main` branch; its default is `v2`, and this work sits on top of `2.1`. Targeting `v2` would have bundled 8 unrelated commits into the diff.
- **`--base 2.1` vs the ballot branch.** If this should instead land on `2.1.0-ballot.2026-07`, retarget — the branch point is the same commit.
- **TASC is internally inconsistent on the role name.** It says *product owners* in every decision-maker list but *product leads* once, in the pre-release sentence. This PR standardizes on *product owners* throughout; may be worth cleaning up upstream.
- **The maturity criteria table is still image-only** (`maturity_levels.png`). Its content matches TASC Table 1 exactly, including the Deprecated level, so there is no divergence — but "Deprecated" appears nowhere in the page's prose and the criteria are not searchable or screen-readable. Converting it to an rst `list-table` is a reasonable follow-up; deliberately out of scope here.
- **Sibling repos.** Cat-VRS and VA-Spec still title theirs "GKS Maturity Model", so the three now disagree. Not touched.
- Sphinx builds clean (no warnings). Verified that the retitle propagates through the `maturity-model` anchor to the releases index and the generated per-class maturity notes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/source/appendices/maturity_model.rst
+++ b/docs/source/appendices/maturity_model.rst
@@ -1,9 +1,19 @@
 .. _maturity-model:
 
-GKM Maturity Model
-!!!!!!!!!!!!!!!!!!
+GA4GH Maturity Model
+!!!!!!!!!!!!!!!!!!!!
 
-The Genomic Knowledge Models work stream is developing semantic data exchange
+.. note::
+
+    This page describes how the GA4GH maturity model is applied to the Genomic
+    Knowledge Models (GKM) — VRS, Cat-VRS, and VA-Spec — developed by the GA4GH
+    Genomic Knowledge Standards (GKS) Work Stream. The maturity model, its release
+    process, and its versioning policy are defined by the GA4GH Technical Alignment
+    Subcommittee (TASC) in the `Technical Specification Development`_ policy, which
+    is authoritative. Where this page adds detail, it does so within the bounds of
+    that policy.
+
+The GA4GH Genomic Knowledge Standards (GKS) Work Stream is developing semantic data exchange
 standards for federated genomic knowledge sharing. To address this, new technical
 specifications are required, such as the VRS standard, which must be developed
 and iterated upon through application across community implementations. This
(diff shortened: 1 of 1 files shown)


larrybabb (approved):
ship it

